PR #79, opened 2021-01-11
Fix 4 small bugs in the contract

### LINKED ISSUE #70: ERC20 transfers tracking error when "from" address equals "to" address
When a particular address performs a transfer to itself the Network Subgraph is not accounting that event properly and the effect is that those tokens are showing under the address balance.

davekaj:
- erc20 transfers to self are now handled for GRT fixed #70
- delegator rewards no longer created when there is 0 in delegators pool. fixes https://github.com/graphprotocol/graph-network-subgraph/issues/71
- all divisions by zero are now protected
- curation.ts was minusing signal from a token amount

deployed to here for testing https://thegraph.com/explorer/subgraph/davekaj/graph-development


package.json
```diff
@@ -14,10 +14,10 @@
...
-    "deploy-mainnet-staging": "yarn && yarn prepare:mainnet && graph deploy graphprotocol/mainnet-staging--ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
+    "deploy-mainnet-staging": "yarn && yarn prepare:mainnet && graph deploy graphprotocol/mainnet-staging --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
     "deploy-mainnet": "yarn && yarn prepare:mainnet && graph deploy graphprotocol/graph-network-mainnet --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
     "deploy-testnet": "yarn && yarn prepare:rinkeby && graph deploy graphprotocol/graph-network-testnet --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
-    "deploy-testing": "yarn && yarn <INSERT_MAINNET_OR_RINKEBY> && graph deploy <INSERT_SUBGRAPH_NAME> --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
+    "deploy-testing": "yarn && yarn prepare:mainnet && graph deploy davekaj/graph-development --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
```
juanmardefago, line 20: Not sure if this change was intended

src/mappings/rewardsManager.ts
```diff
@@ -18,7 +18,13 @@ export function handleRewardsAssigned(event: RewardsAssigned): void {
   let indexerIndexingRewards = event.params.amount
     .times(BigInt.fromI32(indexer.indexingRewardCut))
     .div(BigInt.fromI32(1000000))
-  let delegatorIndexingRewards = event.params.amount.minus(indexerIndexingRewards)
+
+  // If the delegation pool has zero tokens, the contracts don't give away any rewards
+  let delegatorIndexingRewards =
```
juanmardefago, line 23: Should the `indexerIndexingRewards` also be calculated as if it was receiving 100% of the rewards?

davekaj: good catch, need to fix it

package.json
davekaj: nope, i will set it back

juanmardefago (approved):

juanmardefago:
Code looks good now ;) @davekaj

davekaj:
thanks! merging into mainnet-staging now.

PR #80, opened 2021-01-13
Clean up subgraph

davekaj:
This is a clean up of the Schema. What is changed:

- A lot of comments cleaned up
- Some commented out code deleted
- Some fields permanently removed. These are mostly unimplemented fields. I believe for this update to the subgraph we should permanently remove a few non-used fields and not support them going forward. Because I don't think almost anyone will be using these. We can announce the breaking changes in channels. Going forward, we should aim to deprecate fields So that we aren't breaking the code of people depending on this subgraph
- I went through and labeled some fields as either CURRENT or ACCUMULATING. I want to make it clear to people using the subgraph when a value is CURRENT (your actual amount on chain) or ACCUMULATING (the amount you've deposited continuously adding). I have found many people to be confused by this.
- And a few tiny bug fixes


schema.graphql
```diff
@@ -319,22 +316,6 @@ type Subgraph @entity {
   website: String!
   "Display name"
   displayName: String!
-
-  # Aggregate data, by combining all versions, and all deployments those versions point to
```
davekaj, line 323: Removed these for now, they were not in use, and a lot more thinking needs to be done if we want to include them

src/mappings/curation.ts
```diff
@@ -103,10 +103,7 @@ export function handleBurned(event: Burned): void {
   deployment.signalAmount = deployment.signalAmount.minus(event.params.signal)
   deployment.save()
 
-  // Update epoch
-  let epoch = createOrLoadEpoch(event.block.number)
```
davekaj, line 107: Removed unsignaled tokens. the field should just accumulate signaled tokens in an epoch

schema.graphql
```diff
@@ -34,28 +34,26 @@ type GraphNetwork @entity {
...
   channelDisputeEpochs: Int!
-  "Period to wait before delegators can settle"
+  "Epochs to wait before delegators can settle"
   maxAllocationEpochs: Int!
   "Time in blocks needed to wait to unstake"
   thawingPeriod: Int!
   "Minimum time an Indexer must use for resetting their Delegation parameters"
   delegationParametersCooldown: Int!
-  "The number of times indexing rewards are handed out each epoch"
-  indexingRewardsPerEpoch: Int!
```
davekaj, line 58: This field is very old and no longer in use, it represents an old way the rewards were implemented

schema.graphql
```diff
@@ -525,17 +507,15 @@ State channel allocation representing Indexer-SubgraphDeployment stake
 type Allocation @entity {
   "Channel Address"
   id: ID!
-  "Price of the queries"
-  price: BigInt!
```
davekaj, line 529: Removed as there is no longer a plan to store price on chain or on ipfs

src/mappings/staking.ts
```diff
@@ -459,6 +454,12 @@ export function handleAllocationClosed(event: AllocationClosed): void {
   let pool = createOrLoadPool(event.params.epoch)
   // effective allocation is the value stored in contracts, so we use it here
   pool.allocation = pool.allocation.plus(event.params.effectiveAllocation)
+
+  // We must call the contract directly to see how many fees are getting closed in this
+  // allocation. The event does not emit this information
+  let staking = Staking.bind(event.address)
+  let contractAlloc = staking.getAllocation(event.params.allocationID)
+  pool.totalQueryFees = pool.totalQueryFees.plus(contractAlloc.collectedFees)
```
davekaj, line 462: This was a bug I caught, we should have been doing this before

schema.graphql
```diff
@@ -480,20 +462,12 @@ type Indexer @entity {
   "The total amount of indexing rewards given to delegators"
   delegatorIndexingRewards: BigInt!
 
-  # Capacity data
```
davekaj, line 483: Removed these permanently since they don't work if delegation ratio changes.

schema.graphql
```diff
@@ -480,20 +462,12 @@ type Indexer @entity {
...
-  "Amount of delegated tokens that can be eligible for rewards"
-  delegatedCapacity: BigInt!
-  "Total token capacity = delegatedCapacity + stakedTokens"
-  tokenCapacity: BigInt!
-  "Stake available to earn rewards. tokenCapacity - allocationTokens - lockedTokens"
-  availableStake: BigInt!
```
juanmardefago, line 489: I'm quite certain people are using at least this one, should we still remove it with the `calculateCapacities()` function?

If so, we probably need to announce the change with a big disclaimer

davekaj: Agreed, let's put it in the change log for the big release.

The reason I want to remove it is because I don't see a good way that we can keep this consistent. I think it is very possible we change the delegation capacity within the next 6 months, and this will lead many other people to have bad delegation capacities. 

If we can think of a way we could make this work, I would say let's keep it. But I can think of any :(

davekaj: Okay we will keep it in. We introduce some more fields in a newer PR that also has the same issue.  So might as well keep them both in for now. Plan is to make note of the issue in the repository, and figure out a way in the future to avoid this issue, if possible

davekaj:
These are the breaking changes to the schema:

```
- graphnetwork.indexingRewardsPerEpoch - old, never in use

- subgraph.totalIndexingRewards
- subgraph.totalQueryFeesCollected - removed these two because they are super hard to calculate, and not implemented at all, but are very misleading to all people using the subgraph. it is more helpful to have them removed


- allocation.price - not in use at all

```

PR #84, opened 2021-01-14
Added advanced metrics

juanmardefago:
Added the following "advanced metrics" fields:
* `indexerRewardsOwnGenerationRatio`: Percentage of indexer own rewards received in relation to its own stake. 1 (100%) means that the indexer is receiving the exact amount that is generated by his own stake
* `ownStakeRatio`: Ratio between the amount of the indexers own stake over the total usable stake
* `delegatedStakeRatio`: Ratio between the amount of delegated stake over the total usable stake
* `indexingRewardEffectiveCut`: The percent of indexing rewards generated by the delegated stake that the Indexer keeps for itself
* `indexingRewardOverdelegationDilution`: The percent of reward dilution delegators experience because of overdelegation. Overdelegated stake can't be used to generate rewards but still gets accounted while distributing the generated rewards. This causes dilution of the rewards for the rest of the pool

Their respective formulas are isolated in the `helpers.ts` file.


schema.graphql
```diff
@@ -479,6 +479,8 @@ type Indexer @entity {
   indexerIndexingRewards: BigInt!
   "The total amount of indexing rewards given to delegators"
   delegatorIndexingRewards: BigInt!
+  "Percentage of indexer own rewards received in relation to its own stake. 1 (100%) means that the indexer is receiving the exact amount that is generated by his own stake"
```
davekaj, line 482: seems like first use of indexer should be indexers

schema.graphql
```diff
@@ -493,15 +495,23 @@ type Indexer @entity {
...
   "Total shares of the delegator pool"
   delegatorShares: BigInt!
   "Exchange rate of of tokens received for each share"
   delegationExchangeRate: BigDecimal
-  "The percent the Indexer agrees to share with delegators. In parts per million"
+  "The percent of indexing rewards generated by the total stake that the Indexer keeps for itself. In parts per million"
   indexingRewardCut: Int!
+  "The percent of indexing rewards generated by the delegated stake that the Indexer keeps for itself"
+  indexingRewardEffectiveCut: BigDecimal!
+  "The percent of reward dilution delegators experience because of overdelegation. Overdelegated stake can't be used to generate rewards but still gets accounted while distributing the generated rewards. This causes dilution of the rewards for the rest of the pool."
```
davekaj, line 510: description seems long. maybe `The percent of rewards dilution that delegators experience from overdelegation to the indexer.` 

also i think we might be able to just call it `overDelegationDilution` wdyt?

davekaj (changes requested):
minor changes requested, let's chat tomorrow

src/mappings/helpers.ts
```diff
@@ -577,3 +582,78 @@ function createGraphAccountName(
...
+
+function min(a: BigDecimal, b: BigDecimal): BigDecimal {
+  return a < b ? a : b
+}
+
+function max(a: BigDecimal, b: BigDecimal): BigDecimal {
+  return a > b ? a : b
+}
+
+export function calculateOwnStakeRatio(indexer: Indexer): BigDecimal {
```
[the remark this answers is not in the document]
juanmardefago: Umm weird, I thought it was supported. I'm using Atom and it doesn't highlight the same errors, probably related to a missing syntax highlighter plugin though.

I'll swap the formulas to use the functions instead of the operators.

src/mappings/helpers.ts
```diff
@@ -577,3 +582,78 @@ function createGraphAccountName(
...
+  let graphNetwork = GraphNetwork.load('1')
+  let delegationRatio = BigInt.fromI32(graphNetwork.delegationRatio)
+  let maxPossibleTotalUsable =
+    indexer.stakedTokens.toBigDecimal() +
```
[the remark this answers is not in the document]
juanmardefago: Yeah, I totally forgot about this, it wasn't meant to be like this, but swapping formulas wasn't as easy as I would've liked 😂

src/mappings/helpers.ts
```diff
@@ -577,3 +582,78 @@ function createGraphAccountName(
...
+    indexer.stakedTokens.toBigDecimal() + indexer.delegatedTokens.toBigDecimal()
+  let totalUsable = min(maxPossibleTotalUsable, currentTotalStake)
+  return totalUsable == BigDecimal.fromString('0')
+    ? BigDecimal.fromString('0')
+    : indexer.stakedTokens.toBigDecimal() / totalUsable
+}
+
+export function calculateDelegatedStakeRatio(indexer: Indexer): BigDecimal {
+  let graphNetwork = GraphNetwork.load('1')
+  let delegationRatio = BigInt.fromI32(graphNetwork.delegationRatio)
```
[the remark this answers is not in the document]
juanmardefago: Yeah, the only way to prevent this is to use the "Allocated stake based" formula, but I'm not sure that the resulting effective cut will be relevant for delegators with that formula.

As a point of reference, the allocation based formula would show higher effective cut % depending on the amount of unallocated stake that the indexer has, because the ratios would skyrocket as a result (since total usable, delegated and own stake would be far higher than the allocated stake when you have high unallocated stake amounts).

Although, as you mentioned, it should only really affect over-delegated indexers, and it's updated quite often (unlike the delegation ratio), so it shouldn't display "bad" data often

src/mappings/helpers.ts
```diff
@@ -577,3 +582,78 @@ function createGraphAccountName(
...
+  let maxDelegatedStake = indexer.stakedTokens.toBigDecimal() * delegationRatio.toBigDecimal()
+  return indexer.stakedTokens == BigInt.fromI32(0)
+    ? BigDecimal.fromString('0')
+    : BigDecimal.fromString('1') -
+        maxDelegatedStake / max(maxDelegatedStake, indexer.delegatedTokens.toBigDecimal())
+}
+
+export function updateAdvancedIndexerMetrics(indexer: Indexer): Indexer {
+  indexer.ownStakeRatio = calculateOwnStakeRatio(indexer as Indexer)
+  indexer.delegatedStakeRatio = calculateDelegatedStakeRatio(indexer as Indexer)
```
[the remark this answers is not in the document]
juanmardefago: Not really, it can go over 1 in over-delegated scenarios, since `totalUsableStake` would be lower than `ownStake + delegatedStake`

schema.graphql
juanmardefago: Yeah, could be simplified in both naming and description. I wanted to add as much info as possible since most delegators think these metrics are too complex

src/mappings/helpers.ts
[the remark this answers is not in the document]
davekaj: ok lets keep it then!

schema.graphql
davekaj: yes

schema.graphql
davekaj: How about we keep the long comment, but shorten the name?

src/mappings/helpers.ts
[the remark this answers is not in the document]
davekaj: My gut feeling is we should go with what you have implemented here, get some feedback from users, and see how clear it is to them. And then maybe we can change it. 

I wonder if we can get Nena to use it on the front end, before officially releasing an update to everyone, so that maybe we can get feedback and change it

juanmardefago:
Updated with changes, subgraph indexing here: https://thegraph.com/explorer/subgraph/juanmardefago/dev-subgraph?version=pending

src/mappings/helpers.ts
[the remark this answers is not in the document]
davekaj: We decided to leave it as is for now, the operators can be used. It might just be my editor is being too picky. 

The goal is next week to make the whole repository use the operators, but i want to get my editor working first

davekaj (approved):
New update looks good. Approved. Please double check the final numbers once it finishes syncing, then once confirmed, merge :)

src/mappings/helpers.ts
```diff
@@ -592,32 +592,28 @@ function max(a: BigDecimal, b: BigDecimal): BigDecimal {
...
-    maxPossibleTotalUsable,
-    indexer.stakedTokens.toBigDecimal() + indexer.delegatedTokens.toBigDecimal(),
-  )
+  let delegationRatioBD = BigInt.fromI32(graphNetwork.delegationRatio).toBigDecimal()
+  let maxPossibleTotalUsable = stakedTokensBD + stakedTokensBD * delegationRatioBD
+  let totalUsable = min(maxPossibleTotalUsable, stakedTokensBD + delegatedTokensBD)
   return totalUsable == BigDecimal.fromString('0')
     ? BigDecimal.fromString('0')
-    : indexer.delegatedTokens.toBigDecimal() / totalUsable
+    : delegatedTokensBD / totalUsable
```
davekaj: I am seeing a weird number here for overdelegated scenarions. See for 0xa3276e7ab0a162f6a3b5aa6b3089accbaa65d12e . In the subgraph we have about 99.5% right now but in the google sheet around 94%. 

The formulas are just different, from what I can tell. Maybe I am missing something though. Let me know if this is a bug , or if this was meant to be

davekaj (changes requested):

src/mappings/helpers.ts
juanmardefago: It should be now working as expected on the dev-subgraph deploy

davekaj: looks good!

davekaj (approved):

trader-payne:
Wow, you guys rock!!!

PR #98, opened 2021-02-19
Refactor for delegationExchangeRate and truncation of BD field

juanmardefago:
https://thegraph.com/explorer/subgraph/juanmardefago/dev-subgraph?version=pending

From a quick glance at the current state, there are no instances of negative realized rewards, and exchange rates truncated line up perfectly


src/mappings/helpers.ts
```diff
@@ -111,7 +111,7 @@ export function createOrLoadIndexer(id: string, timestamp: BigInt): Indexer {
     indexer.ownStakeRatio = BigDecimal.fromString('0')
     indexer.delegatedStakeRatio = BigDecimal.fromString('0')
     indexer.delegatorShares = BigInt.fromI32(0)
-    indexer.delegationExchangeRate = BigDecimal.fromString('0')
+    indexer.delegationExchangeRate = BigDecimal.fromString('1')
```
davekaj, line 114: im sure this is okay, but just curious, what made you change it to 1?

src/mappings/helpers.ts
```diff
@@ -183,7 +183,7 @@ export function createOrLoadDelegatedStake(
     delegatedStake.lockedTokens = BigInt.fromI32(0)
     delegatedStake.lockedUntil = 0
     delegatedStake.shareAmount = BigInt.fromI32(0)
-    delegatedStake.personalExchangeRate = BigDecimal.fromString('0')
+    delegatedStake.personalExchangeRate = BigDecimal.fromString('1')
```
davekaj, line 186: same question as above

davekaj (approved):
looks good and I also looked at the subgraph results in detail and they look good

src/mappings/helpers.ts
juanmardefago: Both for consistency, since `delegationExchangeRate` always starts at 1 when the first delegations flow into the indexer.
Both values are overwritten by their corresponding calculations on stake delegation though, so it's mostly to keep consistency of the values, since they can never be actually 0.

PR #100, opened 2021-03-18
Add automated deploy to staging

nenadjaja:


juanmardefago (commented):
Code seems good, checked against our npm scripts and it matches up.

PR #101, opened 2021-03-24
Fix epoch calculation when epoch length was changed

abarmat:


juanmardefago (approved):
Code looks good

juanmardefago (commented):
looks good, just 1 doubt, what would happen if the newEndBlock has already passed?
Should we create a new epoch? (or multiple new epochs if more than one full epoch is missing?)

abarmat:
> @abarmat looks good, just 1 doubt, what would happen if the newEndBlock has already passed?
> Should we create a new epoch? (or multiple new epochs if more than one full epoch is missing?)

The added code is only to ensure that the latest saved epoch has the right endBlock matching the epoch length. The edge case is when there is a running epoch already stored and we extended the epoch length.

We discussed with @davekaj once about filling the missing epochs or not. I would do it only if we need to. 
I'm thinking about the condition when multiple empty epochs pass and then epoch length is changed.

davekaj:
Seems like in reality only Epoch 2 will ever be missing. and that this should be a backlog fix as its very low priority (unless we start doing data analysis? would help with empty data, when not expected)

abarmat:
Yes, I don't think it is high priority

PR #107, opened 2021-04-07
Staging to Production PR

juanmardefago:


src/mappings/epochManager.ts
```diff
@@ -38,25 +38,22 @@ export function handleEpochLengthUpdate(event: EpochLengthUpdate): void {
...
   let epoch = createOrLoadEpoch(event.block.number)
-  // We must take this epoch, update its blocks and end it, and make a new epoch
-  // Since whenever we update epoch length in the contracts we end the current epoch
-  // and create a new one
-  epoch.endBlock = event.block.number.toI32()
-  epoch.save()
+
+  // Check that the endBlock for the current epoch match what it should based on the
+  // changed epoch length
+  let newEndBlock = epoch.startBlock + event.params.epochLength.toI32()
```
davekaj, line 44: this part confuses me a bit, what are we doing exactly?

davekaj (approved):

src/mappings/epochManager.ts
juanmardefago: @abarmat I think it's related to your epoch fix?

src/mappings/epochManager.ts
```diff
@@ -38,25 +38,22 @@ export function handleEpochLengthUpdate(event: EpochLengthUpdate): void {
...
+  if (epoch.endBlock != newEndBlock) {
```
davekaj, line 45: Specically this line is confusing me

my interpretation is that `if (epoch.endBlock != newEndBlock)`  should be removed

the way it is implemented right now it looks like to me that, in the following scenario:

```
- epoch start block = 1000
- epoch end block = 3000
then we halve the epoch length. so now it will be
  let newEndBlock = 1000 + 1000
  if (3000 != 2000) {
    epoch.endBlock = 2000
    epoch.save()
  }
but if we double the epoch length we get
  let newEndBlock = 1000 + 4000
  if (3000 != 5000) {
    epoch.endBlock = 5000
    epoch.save()
  }
  ```

it seems in both cases we want it to update. In the first case, that epoch has effectively ended. and a new epoch will be created on the next event handler where createOrLoadEpoch() is used. in the second case, we are fine, the epoch that exists has been lengthened
The only time the if statement will be false is if the epoch length is changed to its same length. but the contracts have a require making that impossible. therefore it seems to me the if statement has no use
what do you think? or am i missing something ? :smile: @abarmat

src/mappings/epochManager.ts
abarmat: That code is for this fix https://github.com/graphprotocol/graph-network-subgraph/pull/101

Originally the `endBlock` for the current epoch was being set to `block.number` while the real endBlock is when the epoch ends, that is startBlock + epochLength

abarmat: That we creating a shorter epoch when epoch length was changed

src/mappings/epochManager.ts
abarmat: Ok, we can remove it, I just included that clause to avoid the extra load of saving the entity again if not needed, but as you say, the handler will only be called when a different length was used

PR #113, opened 2021-04-29
Fix curation tax missing from Signaled event

davekaj:
Pretty straight forward. Also i deleted `deployment.json` since that is an old file not in use anymore

Its syncing here https://thegraph.com/explorer/subgraph/davekaj/dave-graph-mainnet?version=pending

we test by checking the `subgraphDeployment.signalledTokens` matches `getCurationPoolTokens()` here https://etherscan.io/address/0x8fe00a685bcb3b2cc296ff6ffeab10aca4ce1538#readProxyContract

Before,  `subgraphDeployment.signalledTokens` would be 2.5% greater (as long as no one has burned any signal. i used omen subgraph and it worked for the quick check)


juanmardefago (approved):

juanmardefago:
@davekaj can I merge this now or do you need to check something else?

PR #114, opened 2021-04-30
Average cost basis for Signal (and Curator)

juanmardefago:
Indexing here: https://thegraph.com/explorer/subgraph/juanmardefago/dev-subgraph2?version=pending


davekaj (approved):
i looked over it in detail, all looks good.

when you get a chance to double check the results, and they look good, we can merge :)

PR #118, opened 2021-05-07
Ipfs "feature flag"

juanmardefago:
* Added `prep:ipfs` and `prep:no-ipfs` to the package json
* Added template files for GNS and DID handlers
* `prepare:mainnet` and `prepare:rinkeby` by default use `prep:ipfs`
* Made optional some remaining IPFS related fields

This shouldn't impact our deployments to mainnet-staging (since I didn't remove the final GNS and DID files for IPFS), albeit we should update our github actions to also include prepping those two template files, and later down the road remove the final files (gns.ts, and the DID registry one) from this repo, so only the templates are modified.


davekaj:
i will test this tomorrow

davekaj (approved):
It looks good to me! I ran the script and it worked and the code looks clean. I have two suggestions:

- I think it would be good to have something like `deploy-mainnet-ipfs` and `deploy-mainnet-noipfs` as scripts. 
- Did we test to see that the non-ipfs one syncs and works?

juanmardefago:
> It looks good to me! I ran the script and it worked and the code looks clean. I have two suggestions:
> 
>     * I think it would be good to have something like `deploy-mainnet-ipfs` and `deploy-mainnet-noipfs` as scripts.
> 
>     * Did we test to see that the non-ipfs one syncs and works?

Agree, although I didn't add them because I wanted to check whether the solution looked clean before that, will add that now.

I did test that it synced a bit, but since I was working in another task I overwrote the subgraph before it finished syncing, so I'll redeploy a non-ipfs version to another subgraph and let it fully sync

davekaj:
sounds good, once both are done lets merge

juanmardefago:
@davekaj had to do some git magic to fix this branch, since for some reason I managed to get some commits mixed up and it couldn't be automatically merged to mainnet-staging.

I'll redeploy to make sure everything worked as expected and get back to this.

src/mappings/ethereumDIDRegistry.template.ts
```diff
@@ -0,0 +1,39 @@
...
+      graphAccount.description = jsonToString(data.get('description'))
+      graphAccount.image = jsonToString(data.get('image'))
+      graphAccount.displayName = jsonToString(data.get('displayName'))
+      let isOrganization = data.get('isOrganization')
+      if (isOrganization != null && isOrganization.kind === JSONValueKind.BOOL) {
+        graphAccount.isOrganization = isOrganization.toBool()
+      }
+      graphAccount.website = jsonToString(data.get('website'))
+    }
+    {{/ipfs}}
```
Jannis, line 36: I would do this differently. What I would do is add a file called `ipfsHelpers.ts.template` and put the following functions in there:

```
export function fetchGraphAccountMetadata(graphAccount: GraphAccount, ipfsHash: Bytes): void {
  {{#ipfs}}
  ...
  {{/ipfs}}
  // otherwise do nothing
}

export fetchSubgraphMetadata(subgraph: Subgraph, ipfsHash: Bytes): void {
  {{#ipfs}}
  ...
  {{/ipfs}}
  graphAccount.description = ''
  ...
}
```

This way the templating is limited to only the IPFS-related code and the `gns.ts` mapping is unaffected. Feels cleaner and more isolated to me.

Jannis (changes requested):

juanmardefago:
Closing since it's been already re-done by @Jannis on #126

PR #126, opened 2021-05-22
IPFS feature flag v2

Jannis:
Slightly better separation of the IPFS related code into a `metadataHelpers.ts` template, compared to #118.

Also fixed the IPFS node for the testnet. We're using `ipfs.network.thegraph.com` for both networks.


davekaj (approved):
lgtm, @Jannis did you get the subgraph syncing somewhere and confirm its working as intended? we always check that before merging

If so I think we are good to go, unless @juanmardefago has any concerns

Jannis:
> lgtm, @Jannis did you get the subgraph syncing somewhere and confirm its working as intended? we always check that before merging
> 
> If so I think we are good to go, unless @juanmardefago has any concerns

Yes:

- https://thegraph.com/explorer/subgraph/jannis/graph-network-testnet-no-ipfs
- https://thegraph.com/explorer/subgraph/jannis/graph-network-mainnet-no-ipfs

src/mappings/gns.ts
```diff
@@ -80,25 +81,11 @@ export function handleSubgraphMetadataUpdated(event: SubgraphMetadataUpdated): v
...
-    } else {
-      subgraph.description = ''
-      subgraph.image = ''
-      subgraph.displayName = ''
-      subgraph.codeRepository = ''
-      subgraph.website = ''
-    }
-  }
+
+  fetchSubgraphMetadata(subgraph, base58Hash)
```
juanmardefago, line 87: @davekaj Are we 100% sure that passing the entity and updating it in another function (without returning it) works as expected?

We usually did this and returned the resulting entity so that we could replace the variable pointer afterwards to make sure the `.save()` function actually saved the changes we did on the auxiliary function, but I can't remember if that was because of some bug or because of how reference/pointers worked in `AssemblyScript` 🤔 

Also, if I recall correctly, the build stage usually whines if we pass an entity type parameter without explicitly casting it (`subgraph as Subgraph`).

src/mappings/gns.ts
```diff
@@ -160,19 +147,10 @@ export function handleSubgraphPublished(event: SubgraphPublished): void {
...
-      let data = tryData.value.toObject()
-      subgraphVersion.description = jsonToString(data.get('description'))
-      subgraphVersion.label = jsonToString(data.get('label'))
-    } else {
-      subgraphVersion.description = ''
-      subgraphVersion.label = ''
-    }
-  }
+
+  fetchSubgraphVersionMetadata(subgraphVersion, base58Hash)
```
juanmardefago, line 152: Same as above

src/mappings/gns.ts
juanmardefago: Also, this possible issue wouldn't be spotted on a non-ipfs deployment, since the issue would be that the changes within that function (if any) wouldn't be actually stored in the entity after the function execution.

Jannis: I'm pretty sure this works but I'd be happy to return the subgraph entity again. I'll make that change.

Jannis:
@juanmardefago Addressed your comments and tested building with and without the IPFS feature.

juanmardefago (approved):

juanmardefago:
@Jannis yeah, sorry for the nitpick, I'll try to test the details on how object reference and scoping works on cases like this for the future. LGTM!

src/mappings/gns.ts
davekaj: The one thing i am unsure of is :

- if you pass the entity to `function x()`, `save()` it  in there, and don't return it, and continue forward in the original `function y()`, then the entity in `y()` will not have the updated fields. If you then save again, you overwrite the changes that were made in `x()`. 

I _think_ that might happen. But either way, Jannis' fix means we don't have to worry if this is exactly how it works

PR #128, opened 2021-06-08
Added Networks and Categories

juanmardefago:
Addition to the `Subgraph` entity. Read the data from the same IPFS file as the rest of the metadata for the subgraphs, so it's only gonna be enabled for IPFS-enabled deployments


nenadjaja (approved):
Looks good to me! +1

PR #131, opened 2021-06-15
NameSignal refactor and a few improvements

juanmardefago:
* Added `signal` values to the `NameSignal` entity. This was done so that the `Curator` entity could then have an accurate depiction of the total average cost basis and total signal values.
* Added all related `averageCostBasis` fields for `signal` on the `NameSignal` entity.
* `totalSignalAverageCostBasis` and `totalSignal` on the `Curator` entity now correctly depicts signals independently of where they where actually signalled from (GNS or Curation contracts). `totalNameSignalAverageCostBasis` and `totalNameSignal` will still be valid and only track those amounts related only to the name signals (GNS).
* Fixed an issue where the `nameSignal` would be 0, but the average cost basis field still had a non-zero value.

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -320,6 +320,8 @@ type Subgraph @entity {
   currentSignalledTokens: BigInt!
   "The CURRENT name signal amount for this subgraph"
   nameSignalAmount: BigInt!
+  "Current amount of version signal managed by the name pool"
+  signalAmount: BigInt!
   "Reserve ratio of the name curation curve. In parts per million"
   reserveRatio: Int!
   "Tokens that can be withdrawn once the Subgraph is deprecated"
--- a/schema.graphql
+++ b/schema.graphql
@@ -330,6 +330,8 @@ type Subgraph @entity {
   withdrawnTokens: BigInt!
   "Curators of this subgraph deployment"
   nameSignals: [NameSignal!]! @derivedFrom(field: "subgraph")
+  "Total amount of NameSignal entities"
+  nameSignalCount: Int!
 
   # Metadata from IPFS linked in GNS
   "Subgraph metadata"
@@ -882,6 +884,18 @@ type NameSignal @entity {
   signalAverageCostBasis: BigDecimal!
   "signalAverageCostBasis / signal"
   signalAverageCostBasisPerSignal: BigDecimal!
+}
+
+"""
+Auxiliary entity to be able to batch update NameSignal entities
+"""
+type NameSignalSubgraphRelation @entity {
+  "Subgraph ID + index"
+  id:ID!
+
+  nameSignal: NameSignal!
+
+  subgraph: Subgraph!
 }
 
 """
--- a/schema.graphql
+++ b/schema.graphql
@@ -865,7 +865,7 @@ type NameSignal @entity {
   "Shares of the name pool (GNS) that the curator has from signaling their GRT"
   nameSignal: BigInt!
   "Actual signal shares that the name pool minted with the GRT provided by the curator"
-  signal: BigInt!
+  signal: BigDecimal!
 
   # Metrics
   "Block for which the curator last entered or exited the curve"
@@ -874,6 +874,11 @@ type NameSignal @entity {
   # unrealized gains for their current balance, based on the time since the last exit/entry of the curve
   "Summation of realized rewards from before the last time the curator entered the curation curve"
   realizedRewards: BigInt!
+
+  "[DEPRECATED] Curator average cost basis for this name signal on this subgraph. New field for further versions will be nameSignalAverageCostBasis"
+  averageCostBasis: BigDecimal! # note this is ONLY name signal. This is okay for the protocol for now
+  "[DEPRECATED] nameSignalAverageCostBasis / nameSignal. New field for further versions will be nameSignalAverageCostBasisPerSignal"
+  averageCostBasisPerSignal: BigDecimal!
 
   "Curator average cost basis for this name signal on this subgraph"
   nameSignalAverageCostBasis: BigDecimal! # note this is ONLY name signal. This is okay for the protocol for now
--- a/schema.graphql
+++ b/schema.graphql
@@ -316,6 +316,8 @@ type Subgraph @entity {
   signalledTokens: BigInt!
   "CUMULATIVE unsignalled tokens on this subgraph all time"
   unsignalledTokens: BigInt!
+  "CURRENT amount of tokens signalled on this subgraph. Only accounts for signalling done through GNS (name signalling)"
+  currentSignalledTokens: BigInt!
   "The CURRENT name signal amount for this subgraph"
   nameSignalAmount: BigInt!
   "Reserve ratio of the name curation curve. In parts per million"
@@ -772,15 +774,16 @@ type Curator @entity {
   totalReturn: BigDecimal!
   "NOT IMPLEMENTED - Signaling efficiency of the curator"
   signalingEfficiency: BigDecimal!
+
   "CURRENT summed name signal for all bonding curves"
   totalNameSignal: BigDecimal!
-  "Total curator cost basis of all shares purchased on all bonding curves"
+  "Total curator cost basis of all shares of name pools purchased on all bonding curves"
   totalNameSignalAverageCostBasis: BigDecimal!
   "totalNameSignalAverageCostBasis / totalNameSignal"
   totalAverageCostBasisPerNameSignal: BigDecimal!
   "CURRENT summed signal for all bonding curves"
   totalSignal: BigDecimal!
-  "Total curator cost basis of all shares purchased on all bonding curves"
+  "Total curator cost basis of all version signal shares purchased on all bonding curves. Includes those purchased through GNS name pools"
   totalSignalAverageCostBasis: BigDecimal!
   "totalSignalAverageCostBasis / totalSignal"
   totalAverageCostBasisPerSignal: BigDecimal!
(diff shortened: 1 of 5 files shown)


schema.graphql
```diff
@@ -813,10 +820,10 @@ type NameSignal @entity {
   unsignalledTokens: BigInt!
   "Tokens the curator has withdrawn from a deprecated name curve"
   withdrawnTokens: BigInt!
-  "Signal that the curator has from signaling their GRT"
+  "Shares of the name pool (GNS) that the curator has from signaling their GRT"
   nameSignal: BigInt!
-  # Note , we don't show vSignal here. We could, but it has almost no value to display it.
-  # We can get away with using curator nameSignal, total n and v signal on the subgraph.
+  "Actual signal shares that the name pool minted with the GRT provided by the curator"
+  signal: BigInt!
```
davekaj, line 826: this is another DYNAMIC value. if someone else starts minting alongside the GNS signal/vSignal, they could end up with negative signal here. or could sell all their nameSignal and still have signal here, even though thats not true. 

how do we plan to use this?

schema.graphql
```diff
@@ -825,10 +832,16 @@ type NameSignal @entity {
...
   "Curator average cost basis for this name signal on this subgraph"
-  averageCostBasis: BigDecimal! # note this is ONLY name signal. This is okay for the protocol for now
-  "averageCostBasis / nameSignal"
-  averageCostBasisPerSignal: BigDecimal!
+  nameSignalAverageCostBasis: BigDecimal! # note this is ONLY name signal. This is okay for the protocol for now
+  "nameSignalAverageCostBasis / nameSignal"
+  nameSignalAverageCostBasisPerSignal: BigDecimal!
+
+  "Curator average cost basis for the version signal on this subgraph name pool"
+  signalAverageCostBasis: BigDecimal!
```
davekaj, line 842: this would be effected by the comment above, where i mention that `namesignal.signal` is dynamic.

davekaj (changes requested):
A few questions to discuss, we can do here or over slack

schema.graphql
```diff
@@ -299,6 +303,8 @@ type Subgraph @entity {
   signalledTokens: BigInt!
   "CUMULATIVE unsignalled tokens on this subgraph all time"
   unsignalledTokens: BigInt!
+  "CURRENT amount of tokens signalled on this subgraph. Only accounts for signalling done through GNS (name signalling)"
+  currentSignalledTokens: BigInt!
```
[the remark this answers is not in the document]
juanmardefago: Good point, the main idea is that the frontend needs to have a way of displaying the current amount of GRT signalled for a subgraph, and also have that value be sortable (so it needs to be a specific field, and it needs to be on the `Subgraph` entity).

Since the `Subgraph` entity representation of signalling is `NameSignal` (given the long discussion we had about signal vs nameSignal, double counting curation and what is a subgraph), and given it had 2 cumulative counters (`signalledTokens` and `unsignalledTokens`) I basically replicated the `signalledTokens - unsignalledTokens` value on a new separate field, which should ideally be the amount of GRT signalled through name signals, but now that you mention it, unsignalled can be higher if people curate outside of the GNS name signal system, and someone from the GNS name signal system unsignals...

All I can think as a better solution is to directly link the latest `SubgraphDeployment` entity, and have this new field track the token changes with version signals instead 🤔, or something along those lines

schema.graphql
juanmardefago: This was mostly geared towards being able to track `totalSignal` and `totalAverageCostBasis` on the `Curator` entity, since what we had before were 2 different fields for `totalAverageCostBasis` and `totalSignal`, one for the curation done through name signals, and one for the curation done through version signals.

The idea was to get the actual amount of vSignal minted each time someone signalled through GNS, so that we can use the actual version signal to aggregate towards the `totalAverageCostBasis` and `totalSignal` field of the `Curator` for the version signal. (since 1 vSignal != 1nSignal).

This was intended to be able to have a "Total signal" and "total GRT on curation" (can't remember how it was named), in the curators table (so the totals were the totals for a given Curator/GraphAccount, across all the subgraphs)

schema.graphql
juanmardefago: Somewhat related to the answer above.

schema.graphql
davekaj: i believe it would be fine until there was a combination of :

- someone signals directly on the subgraph deployment, outside of the GNS
- the owner upgrades. then the signals will be changed

I think, then , how to fix it after this scenario, would be a batch update to take the new ratio of `nSignal to vSiganl` and apply it to every curator. But it might get a bit funky since `totalSignal` is an aggregation of many `signals` from many subgraphs

juanmardefago: After revisiting all cases I can think of, I agree that the only way this signal amount would be incorrect is in cases where a GNS signal migration occurs (because of an upgrade) and only if there's more GRT in the curve that was added directly through the curation contracts.

That would make the GNS unsignalling return more GRT than the original amount (since it basically would "rug pull" those that came after GNS), and thus, when signalling on the new version, the amount of version signal minted would be greater, but the amount of name signal would still be the same.

The fix would be the signals ratio change and batch update (as you mentioned), and I think it can be done safely since we can subtract the old version signal value that we had on the `NameSignal` entity, from the `Curator` total signal, and then reapply the new value we calculate based on the new nSignal/vSignal ratio.

davekaj: Yes I think it can be done as well. if you are up for it, lets try it!

schema.graphql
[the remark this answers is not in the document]
juanmardefago: Gonna leave this implemented like this, so it will still have this "known issue" where it can be negative on full unsignals.

It's at least better than have it be a cumulative value used as a current value, and the other alternatives will require something like we did for delegators in the analytics subgraph.

schema.graphql
juanmardefago: This should be fixed with the new changes, following the solution described above.

schema.graphql
juanmardefago: Same as above, should be fixed now

src/mappings/gns.ts
```diff
@@ -328,9 +394,54 @@ export function handleNameSignalUpgrade(event: NameSignalUpgrade): void {
...
+    let nameSignal = NameSignal.load(relation.nameSignal)
+    if(!nameSignal.nameSignal.isZero()) {
+      let curator = Curator.load(nameSignal.curator)
+
+      let oldSignal = nameSignal.signal;
+      nameSignal.signal = nameSignal.nameSignal * signalRatio
+
+      // zero division protection
+      if (nameSignal.signal.toBigDecimal() != zeroBD) {
+        nameSignal.signalAverageCostBasisPerSignal = nameSignal.signalAverageCostBasis.div(
```
davekaj, line 416: does this need truncate?

src/mappings/gns.ts
```diff
@@ -328,9 +394,54 @@ export function handleNameSignalUpgrade(event: NameSignalUpgrade): void {
...
+      if (nameSignal.signalAverageCostBasis == zeroBD) {
+        nameSignal.signalAverageCostBasisPerSignal = zeroBD
+      }
+
+      curator.totalSignal = curator.totalSignal.minus(oldSignal.toBigDecimal()).plus(nameSignal.signal.toBigDecimal())
+      curator.totalSignalAverageCostBasis = curator.totalSignalAverageCostBasis.minus(diffACBSignal)
+      if (curator.totalSignal == zeroBD) {
+        curator.totalAverageCostBasisPerSignal = zeroBD
+      } else {
+        curator.totalAverageCostBasisPerSignal = curator.totalSignalAverageCostBasis.div(
```
davekaj, line 437: does it need truncate?

src/mappings/gns.ts
```diff
@@ -328,9 +394,54 @@ export function handleNameSignalUpgrade(event: NameSignalUpgrade): void {
   // Weirdly here, we add the token amount to both, but also the name curator owner must
   // stake the withdrawal fees, so both balance fairly
   // TODO - will have to come back here and make sure my thinking is correct
+  // event.params.newVSignalCreated -> will be used to calculate new nSignal/vSignal ratio
+  subgraph.signalAmount = event.params.newVSignalCreated
   subgraph.unsignalledTokens = subgraph.unsignalledTokens.plus(event.params.tokensSignalled)
   subgraph.signalledTokens = subgraph.signalledTokens.plus(event.params.tokensSignalled)
   subgraph.save()
+
+  let signalRatio = subgraph.signalAmount / subgraph.nameSignalAmount
```
davekaj, line 403: I believe we will have to update both of these types to BigDecimal. We can get in a scenario where the new signal amount is less than the name signal amount. If that happened, division of two big ints would just give us zero

davekaj (changes requested):
three similar comments about truncate - which i am unsure if it is needed. 

one comment about changing the units so that `signalRatio` will work.

src/mappings/curation.ts
```diff
@@ -155,6 +155,7 @@ export function handleBurned(event: Burned): void {
   signal.averageCostBasis = signal.signal
     .toBigDecimal()
     .times(signal.averageCostBasisPerSignal)
+    .truncate(18)
```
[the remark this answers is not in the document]
juanmardefago: Might be, although I only truncated those that I was 100% sure needed truncation, that's why I didn't do it on all of them 😅

src/mappings/gns.ts
juanmardefago: Hmm, you are right, I forgot the `/` returned a BigInt 🤦🏼

src/mappings/curation.ts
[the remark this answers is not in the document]
davekaj: im okay either way, i was just pointing them out in case they were accidentally forgotten :)

PR #132, opened 2021-06-18
Price per share fix with a couple other small changes

juanmardefago:
Splitted the part that didn't need changes of #131 so we can merge it sooner.
Also kept `pastVersions` to avoid any breaking changes

### DIFF
--- a/src/mappings/helpers.ts
+++ b/src/mappings/helpers.ts
@@ -59,6 +59,7 @@ export function createOrLoadSubgraphDeployment(
   subgraphID: string,
   timestamp: BigInt,
 ): SubgraphDeployment {
+  let graphNetwork = GraphNetwork.load('1')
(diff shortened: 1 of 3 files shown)


src/mappings/helpers.ts
```diff
@@ -676,45 +683,58 @@ export function updateDelegationExchangeRate(indexer: Indexer): Indexer {
 }
 
 export function calculatePricePerShare(deployment: SubgraphDeployment): BigDecimal {
+  // Ideally this would be a 1 to 1 recreation of the share sell formula, but due to
```
davekaj, line 686: Please add a TODO here with what we talked about over slack. We should come back to this when we have time and figure out why there is a slight deviation

davekaj (approved):
one comment, just add a TODO explaining we will fix in the near future

PR #133, opened 2021-06-19
Added active delegator count, as well as delegations count

juanmardefago:
Added a few extra counters for active and total delegations (DelgatedStake), as well as active delegators

currently indexing in a dev subgraph: https://thegraph.com/explorer/subgraph/juanmardefago/dev-subgraph2?version=pending

Gonna be testing the resulting numbers afterwards, but should be accurate.

A delegator/delegation is considered inactive as soon as shares go to zero, and active when shares are higher than zero.

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -155,8 +155,14 @@ type GraphNetwork @entity {
   indexerCount: Int!
   "Number of indexers that currently have some stake in the protocol"
   stakedIndexersCount: Int!
-  "Total delegators"
+  "Total amount of delegators historically"
   delegatorCount: Int!
+  "Total active delegators. Those that still have at least one active delegation."
+  activeDelegatorCount: Int!
+  "Total amount of delegations historically"
+  delegationCount: Int!
+  "Total active delegations. Those delegations that still have GRT staked towards an indexer"
+  activeDelegationCount: Int!
   "Total curators"
   curatorCount: Int!
   "Total subgraphs"
@@ -681,6 +687,11 @@ type Delegator @entity {
   createdAt: Int!
   "Total realized rewards on all delegated stakes. Realized rewards are added when undelegating and realizing a profit"
   totalRealizedRewards: BigDecimal!
+
+  "Total DelegatedStake entity count (Active and inactive)"
+  stakesCount: Int!
+  "Active DelegatedStake entity count. Active means it still GRT delegated"
+  activeStakesCount: Int!
 }
 
 """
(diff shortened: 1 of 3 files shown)


src/mappings/staking.ts
```diff
@@ -208,15 +208,33 @@ export function handleStakeDelegated(event: StakeDelegated): void {
...
 
+  // reload delegator to avoid edge case where we can overwrite stakesCount if stake is new
+  delegator = Delegator.load(delegatorID) as Delegator
+
+  let isDelegatorBecomingActive = delegator.activeStakesCount == 0 && isStakeBecomingActive
+
   // upgrade graph network
   let graphNetwork = GraphNetwork.load('1')
   graphNetwork.totalDelegatedTokens = graphNetwork.totalDelegatedTokens.plus(event.params.tokens)
+
```
davekaj, line 226: I think this is a bit cleaner :

```
   // DELETE isDelegatorBecomingActive

  // upgrade graph network
  let graphNetwork = GraphNetwork.load('1')
  graphNetwork.totalDelegatedTokens = graphNetwork.totalDelegatedTokens.plus(event.params.tokens)

  if(isStakeBecomingActive) {
    graphNetwork.activeDelegationCount = graphNetwork.activeDelegationCount + 1
    delegator.activeStakesCount = delegator.activeStakesCount + 1
      if( delegator.activeStakesCount == 0) {
          graphNetwork.activeDelegatorCount = graphNetwork.activeDelegatorCount + 1
     }
  }
```
Reason being is that `isDelegatorBecomingActive` requires that `isStakeBecomingActive` is already true. Thus we are checking that `isStakeBecomingActive` is true twice. In what I laid out above, we remove the duplication of this check

davekaj (changes requested):
Just two comments

schema.graphql
```diff
@@ -681,6 +687,11 @@ type Delegator @entity {
   createdAt: Int!
   "Total realized rewards on all delegated stakes. Realized rewards are added when undelegating and realizing a profit"
   totalRealizedRewards: BigDecimal!
+
+  "Total DelegatedStake entity count (Active and inactive)"
+  stakesCount: Int!
+  "Active DelegatedStake entity count. Active means it still GRT delegated"
```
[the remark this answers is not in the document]
juanmardefago: Whoops, fixing it

src/mappings/staking.ts
juanmardefago: yeah, that would be cleaner, although I added both as independent variables so that it was easier to read. I'm gonna fix the missing `has` on the previous comment, so I can change this too if you'd like

juanmardefago:
@davekaj changes pushed, should be cleaner and still working as expected!

nenadjaja (approved):
Looks good to me, but maybe @davekaj  wants to double check

davekaj (approved):

PR #137, opened 2021-06-23
Add in token lock wallets

davekaj:
syncing here https://thegraph.com/explorer/subgraph/davekaj/graph-network-combine

### DIFF
--- a/config/mainnetAddressScript.ts
+++ b/config/mainnetAddressScript.ts
@@ -21,6 +21,7 @@ export let addresses: Addresses = {
   ensPublicResolver: '{{mainnet.IPublicResolver}}',
   blockNumber: '',
   network: '',
+  tokenLockManager: ''
 }
 
 const main = (): void => {
@@ -28,6 +29,7 @@ const main = (): void => {
     let output = JSON.parse(mustache.render(JSON.stringify(addresses), renameAddresses))
     output.blockNumber = '11440000' // Hardcoded a few thousand blocks before 1st contract deployed
     output.network = 'mainnet'
+    output.tokenLockManager = "0xFCf78AC094288D7200cfdB367A8CD07108dFa128"
     fs.writeFileSync(__dirname + '/generatedAddresses.json', JSON.stringify(output, null, 2))
   } catch (e) {
     console.log(`Error saving artifacts: ${e.message}`)
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "deploy-mainnet-no-ipfs": "yarn && yarn prep:no-ipfs && yarn prepare:mainnet && graph deploy graphprotocol/graph-network-mainnet --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
     "deploy-testnet": "yarn && yarn prep:ipfs && yarn prepare:rinkeby && graph deploy graphprotocol/graph-network-testnet --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
     "deploy-testnet-no-ipfs": "yarn && yarn prep:no-ipfs && yarn prepare:rinkeby && graph deploy graphprotocol/graph-network-testnet --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
-    "deploy-testing": "yarn && yarn prep:<IPFS_OR_NO-IPFS> && yarn <INSERT_MAINNET_OR_RINKEBY> && graph deploy <INSERT_SUBGRAPH_NAME> --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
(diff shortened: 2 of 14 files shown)


juanmardefago (approved):
Just some minor comments

package.json
```diff
@@ -17,7 +17,7 @@
     "deploy-mainnet-no-ipfs": "yarn && yarn prep:no-ipfs && yarn prepare:mainnet && graph deploy graphprotocol/graph-network-mainnet --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
     "deploy-testnet": "yarn && yarn prep:ipfs && yarn prepare:rinkeby && graph deploy graphprotocol/graph-network-testnet --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
     "deploy-testnet-no-ipfs": "yarn && yarn prep:no-ipfs && yarn prepare:rinkeby && graph deploy graphprotocol/graph-network-testnet --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
-    "deploy-testing": "yarn && yarn prep:<IPFS_OR_NO-IPFS> && yarn <INSERT_MAINNET_OR_RINKEBY> && graph deploy <INSERT_SUBGRAPH_NAME> --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
+    "deploy-testing": "yarn && yarn prep:no-ipfs && yarn prepare:mainnet && graph deploy davekaj/graph-network-combine --ipfs https://api.thegraph.com/ipfs/ --node https://api.thegraph.com/deploy/",
```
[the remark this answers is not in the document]
davekaj: yes, unintended! 😅  , will fix it

src/mappings/metadataHelpers.ts
```diff
@@ -4,67 +4,12 @@ import { jsonToString } from './utils'
 import { createOrLoadSubgraphCategory, createOrLoadSubgraphCategoryRelation, createOrLoadNetwork } from './helpers'
 
 export function fetchGraphAccountMetadata(graphAccount: GraphAccount, ipfsHash: string): void {
-  let ipfsData = ipfs.cat(ipfsHash)
```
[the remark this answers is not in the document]
davekaj: yeah i was wondering what was happening here. I noticed they were deleted, and i didnt delete it, so i figured it was automated. 

I am down to add it to the gitignore in this PR

davekaj:
Ready for review again. we will have to confirm it works when we sync here https://thegraph.com/explorer/subgraph/davekaj/graph-network-combine?version=pending

config/mainnetAddressScript.ts
```diff
@@ -21,13 +21,15 @@ export let addresses: Addresses = {
...
   network: '',
+  tokenLockManager: ''
 }
 
 const main = (): void => {
   try {
     let output = JSON.parse(mustache.render(JSON.stringify(addresses), renameAddresses))
     output.blockNumber = '11440000' // Hardcoded a few thousand blocks before 1st contract deployed
     output.network = 'mainnet'
+    output.tokenLockManager = "0xFCf78AC094288D7200cfdB367A8CD07108dFa128"
```
juanmardefago, line 32: Do we need to do this because the addresses aren't configured in the package?

src/mappings/gns.ts
```diff
@@ -72,10 +73,26 @@ export function handleSetDefaultName(event: SetDefaultName): void {
...
+    addDefaultNameTokenLockWallets(graphAccount, newDefaultName)
   }
   graphAccount.save()
 }
 
+// Add in default names to a graph accounts token lock wallets
+function addDefaultNameTokenLockWallets(graphAccount: GraphAccount, name: string): void {
+  let tlws = graphAccount.tokenLockWallets
+  for (let i = 0; i < tlws.length; i++) {
+    let tlw = GraphAccount.load(tlws[i])
```
juanmardefago, line 85: Are we sure this doesn't return null?
When we create the TLWs in `handleTokenLockCreated` we aren't creating a GraphAccount for it (only for its beneficiary), although it can create one by interacting with the protocol at any point (for example, staking to be an indexer), but what happens with those that don't interact at all? 🤔

juanmardefago: Mostly asking because, since we don't check for it to be null, we can make the subgraph fail on `.save()`, and we could easily fix it by checking for null. Although I might be missing something and it could be initializing the `GraphAccount` every time 😅

juanmardefago (commented):
Few comments

config/mainnetAddressScript.ts
davekaj: exactly. there are not part of the core contracts, and thus are not in the package!

src/mappings/gns.ts
davekaj: I think we should create a graph account for the token lock wallet in `handleTokenLockCreated` . I actually thought that was what I did, but I realize that I only did the beneficiary. 

So if we create both in there, then I believe that should satisfy the situation. What do you think? 

(Also we will still confirm when it is synced)

juanmardefago (approved):

PR #141, opened 2021-07-01
Add two different slashing percentages

abarmat:


abarmat:
@davekaj I left an intentional TODO to review together about the contract interface

juanmardefago:
@abarmat is this already on the contracts?

abarmat:
@juanmardefago it is being voted at the Council this week

abarmat:
My question is what if the `slashingPercentage()` existed as part of the interface in the past but at some block it isn't there, how does a subgraph handle that?

juanmardefago:
@abarmat if you removed it from the code, it's not gonna be consumed at all.
In short, if you use the direct view function (for example: `slashingPercentage()`) and it's not implemented in the contracts (doesn't matter at which point, if you query a contract and it can't handle it) it will break the subgraph.

To avoid this, you can use the `try_` function, in this case `try_slashingPercentage()`, which will not fail when it doesn't work, but simply not return any values.

(same for the new functions, which won't be available on the initial blocks prior to this council vote)

abarmat:
In the case of the new functions they will not be triggered because they will only do so when the right event was emitted. ParameterUpdated('idxSlashingPercentage') can only be emitted on the new contract.

`slashingPercentage()` existed in the past version of the contract, is it an issue of the ABI not having it to query for a time travel query?

juanmardefago:
> In the case of the new functions they will not be triggered because they will only do so when the right event was emitted. ParameterUpdated('idxSlashingPercentage') can only be emitted on the new contract.
> 
> `slashingPercentage()` existed in the past version of the contract, is it an issue of the ABI not having it to query for a time travel query?

yes, ideally you would have both, since the same behavior you mentioned before would still apply (it won't trigger the function if the event isn't triggered), and you would use `try_` functions to prevent any unexpected crashes

abarmat:
I think one of the issues is that the autogenerated ABI won't have the slashingPercentage() and we might need to add it manually

juanmardefago:
> I think one of the issues is that the autogenerated ABI won't have the slashingPercentage() and we might need to add it manually

Yeah it's not an issue, it can be manually stitched in, done that many times 😂 

I can take care of it while I wait for my testing to be unblocked, can you create an asana task for it?

abarmat:
Great :) let me know if you want to sync about it before addressing it so we ensure we don't break stuff. The contract is already upgraded in Rinkeby so it's a good test bed

juanmardefago:
@abarmat can you change the PR to point towards `mainnet-staging` rather than `master`?

juanmardefago:
Indexing here: https://thegraph.com/explorer/subgraph/juanmardefago/dev-subgraph2?query=slashing%20percentage&version=pending

abarmat:
done

juanmardefago:
Works as expected, merging

PR #147, opened 2021-07-22
Batch update deployment signalled tokens to Subgraph entity

juanmardefago:
Currently syncing here: https://thegraph.com/legacy-explorer/subgraph/juanmardefago/dev-subgraph?version=pending

Also fixed some inconsistent spacing on some files, and a weird conflict on a file that we stopped versioning a while ago.


src/mappings/gns.ts
```diff
@@ -166,6 +167,7 @@ export function handleSubgraphPublished(event: SubgraphPublished): void {
 
   // Update subgraph
   let subgraph = createOrLoadSubgraph(subgraphID, event.params.graphAccount, event.block.timestamp)
+  let oldVersionID = subgraph.currentVersion
 
   versionID = joinID([subgraphID, subgraph.versionCount.toString()])
   subgraph.currentVersion = versionID
```
nenadjaja, line 173: Should `let oldVersionID = subgraph.currentVersion` line of code be after this line? (After we set `subgraph.currentVersion`) ?

nenadjaja (commented):
Had one question. 
Might want to wait for @davekaj for approval? Unless this is urgent

src/mappings/gns.ts
juanmardefago: We actually want to maintain the previous version id (before we set the new one, even if it's null) that's why I do it before ;)

src/mappings/metadataHelpers.ts
```diff
@@ -4,10 +4,23 @@ import { jsonToString } from './utils'
 import { createOrLoadSubgraphCategory, createOrLoadSubgraphCategoryRelation, createOrLoadNetwork } from './helpers'
 
 export function fetchGraphAccountMetadata(graphAccount: GraphAccount, ipfsHash: string): void {
+  let ipfsData = ipfs.cat(ipfsHash)
```
davekaj, line 7: Hmm, i remember I did a commit where i made this file git ignored. in fact, `.gitignore` contains it right now. 

i think we can just delete this file from the PR

src/mappings/helpers.ts
```diff
@@ -798,3 +807,48 @@ export function createOrLoadSubgraphCategoryRelation(
...
+  newRelationEntity.deployment = newDeployment.id
+  newRelationEntity.subgraph = subgraph.id
+  newRelationEntity.active = true
+  newRelationEntity.save()
+
+  newDeployment.subgraphCount = newDeployment.subgraphCount + 1
+  newDeployment.activeSubgraphCount = newDeployment.activeSubgraphCount + 1
+  newDeployment.save()
+
+  subgraph.currentVersionRelationEntity = newRelationEntity.id
```
davekaj, line 840: What is the reason for including this field on the subgraph? Just curious

src/mappings/gns.ts
```diff
@@ -191,6 +193,18 @@ export function handleSubgraphPublished(event: SubgraphPublished): void {
...
   subgraphVersion = fetchSubgraphVersionMetadata(subgraphVersion, base58Hash)
   subgraphVersion.save()
+
+  let oldDeployment: SubgraphDeployment | null = null
+  if (oldVersionID != null) {
+    let oldVersion = SubgraphVersion.load(oldVersionID)
+    oldDeployment = SubgraphDeployment.load(oldVersion.subgraphDeployment)
+  }
+  // create deployment - named subgraph relationship, and update the old one
+  updateCurrentDeploymentLinks(
```
davekaj, line 203: I think this function, or one similar to it, will have to run on `handleSubgraphDeprecated`. it might be that we just need to lower the active count, and set active to false. wdyt?

davekaj (changes requested):
A few questions from me

src/mappings/metadataHelpers.ts
juanmardefago: Yeah, I'm not 100% sure why it was still being tracked, but since the old version contained conflict strings, I made sure to at least remove that.
I'm gonna go ahead and delete them on the update.

src/mappings/helpers.ts
juanmardefago: So we can update the old deployment active subgraph count when it changes to a new deployment. (that counter basically adds a little bit of extra information so you can know how many subgraphs are actively using that deployment as their current version)

src/mappings/metadataHelpers.ts
davekaj: yeah i think gitignore does a weird thing where if the file EXISTS it doesnt catch it, and will continue to allow you to edit it. but then once you delete it, gitignore works properly.

src/mappings/gns.ts
juanmardefago: 🤔 
maybe, yeah, although technically the subgraph still has the deployment as its' current version, so it might be weird when querying that data 🤔

juanmardefago: Maybe it would be best to track that as a separate count, like having `subgraphCount`, `activeSubgraphCount` and `deprecatedSubgraphCount`, since an active subgraph for the deployment can stop being active because it was upgraded or because it was deprecated

juanmardefago:
@davekaj rebased to fix conflicts, and changed it to implement a different counter when deprecating, while reusing most of the original code.

juanmardefago:
Currently syncing here: https://thegraph.com/legacy-explorer/subgraph/juanmardefago/dev-subgraph?version=pending

davekaj (approved):
lgtm

PR #150, opened 2021-08-13
Added curation totals breakdown

juanmardefago:


juanmardefago:
Numbers don't match up, will need to check when I'm back from vacation

juanmardefago:
I think it should be fixed now, syncing on the same 2 links

juanmardefago:
New changes display non-negative value tracking based on share value (so any new GRT deposited in the bonding curve will be "distributed" to each of the two categories based on the share distribution on that bonding curve appropriately)

PR #162, opened 2021-11-30
Fix SubgraphVersion.version value setting

fordN:
I did some cleanup based on eslint and prettier suggestions, but the main thrust of this PR is to fix the version property value on `SubgraphVersion` entities (see src/mappings/gns.ts lines 200, 208, 209, and 224). Previously SubgraphVersion.version was always being set to 0. 

_Note: I'm currently testing the change locally._

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 build/
 node_modules/
 src/types
+generated/
--- a/src/mappings/gns.ts
+++ b/src/mappings/gns.ts
@@ -1,4 +1,4 @@
-import { BigInt, BigDecimal, Bytes, json } from '@graphprotocol/graph-ts'
+import { BigInt, BigDecimal, Bytes } from '@graphprotocol/graph-ts'
 import {
   SubgraphPublished,
   SubgraphDeprecated,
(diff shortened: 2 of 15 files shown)


.eslintignore
```diff
@@ -1,3 +1,4 @@
 build/
 node_modules/
-src/types
\ No newline at end of file
+src/types
+generated/
```
juanmardefago, line 4: We didn't include generated since the scripts we use in this repo make sure to send the generated files to `src/types`.
It could be included without any issues, just a quick headsup. (also, it's usually a good idea to use those scripts, since this subgraph uses manifest template, and requires mustache to process a few files before getting usable files for deployment).

We could probably improve the dev workflow though.

src/mappings/gns.ts
```diff
@@ -197,15 +197,16 @@ export function handleSubgraphPublished(event: SubgraphPublished): void {
...
 
   // Update subgraph
   let subgraph = createOrLoadSubgraph(subgraphID, event.params.graphAccount, event.block.timestamp)
   let oldVersionID = subgraph.currentVersion
 
   versionID = joinID([subgraphID, subgraph.versionCount.toString()])
   subgraph.currentVersion = versionID
-  subgraph.versionCount = subgraph.versionCount.plus(BigInt.fromI32(1))
+  versionNumber = subgraph.versionCount.plus(BigInt.fromI32(1))
+  subgraph.versionCount = versionNumber
```
juanmardefago, line 209: Isn't this incorrect? `versionNumber` should be the value of `subgraph.versionCount` before incrementing it (since it's what we are using to create the ID which is used in line `221` to create the entity.

fordN: It works either way I think, but your suggestion would make it more consistent (count from 0 instead of from 1.) I'll update.

.eslintignore
fordN: Well, I'm using scripts in the package.json and I have a `generated` directory. 🤷 
Worth figuring out what I'm doing differently there, but for now it's not hurting anything to make this update to .eslintignore.

juanmardefago (approved):

juanmardefago:
LGTM!

PR #182, opened 2022-04-20
Integrate parsing and entity logic for contracts and events

mnlesane:
https://forum.thegraph.com/t/subgraph-contract-lookup/3274/

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -522,6 +522,9 @@ type SubgraphDeployment @entity {
   activeSubgraphCount: Int!
   "Amount of Subgraph entities that were currently using this deployment when they got deprecated"
   deprecatedSubgraphCount: Int!
+
+  "The contract sources in this subgraph deployment's manifest"
+  contracts: [Contract!]!
 }
 
 # TODO - add when we have the ability to parse data sources
@@ -1220,6 +1223,32 @@ enum Revocability {
 }
 
 """
+Events from contract sources in subgraph manifests
+"""
+type ContractEvent @entity {
+  "Address-Event"
+  id: ID!
+  "Event"
+  event: String
+  "The contract this event is associated with"
+  contract: Contract!
+}
+
+"""
+Contracts in subgraph manifests
+"""
+type Contract @entity {
+  "Contract Address"
+  id: ID!
+  "Placeholder for contract name, if it can be determined"
+  name: String
+  "Subgraph deployments which include this contract in their manifests"
+  subgraphDeployments: [SubgraphDeployment!]! @derivedFrom(field: "contracts")
+  "Events associated with this contract"  
+  contractEvents: [ContractEvent!]! @derivedFrom(field: "contract")
+}
+
+"""
 Full test search for displayName and description on the Subgraph Entity
 """
 type _Schema_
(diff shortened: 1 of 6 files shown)


schema.graphql
```diff
@@ -522,6 +522,9 @@ type SubgraphDeployment @entity {
   activeSubgraphCount: Int!
   "Amount of Subgraph entities that were currently using this deployment when they got deprecated"
   deprecatedSubgraphCount: Int!
+
+  "The contract sources in this subgraph deployment's manifest"
+  contracts: [Contract!]!
```
juanmardefago, line 527: I would probably modify this to use a many-to-many relationship auxiliary entity like described [in the docs](https://thegraph.com/docs/en/developer/create-subgraph-hosted/#many-to-many-relationships).
It's unfortunately a little bit extra effort for querying, but a lot better for indexing performance (mostly db size).

src/mappings/gns.ts
```diff
@@ -250,7 +250,10 @@ export function handleSubgraphPublished(event: SubgraphPublished): void {
   // Create subgraph deployment, if needed. Can happen if the deployment has never been staked on
   let subgraphDeploymentID = event.params.subgraphDeploymentID.toHexString()
   let deployment = createOrLoadSubgraphDeployment(subgraphDeploymentID, event.block.timestamp)
-
+  
+  //Associate Contracts and Events with Deployment
+  processManifestForContracts(subgraph,deployment)
```
juanmardefago, line 255: Just curious, is given that we have functions to abstract the notion of getting new metadata from IPFS, what was the reason to directly call this function in the specific event handlers directly, instead of calling it in the function that fetches the metadata? (I'd assume it's a more direct integration, but we would be calling `ipfs.cat` twice, although not sure how much would that affect performance, and I also think it might be because you wanted to avoid having to change the interface of those fetch functions)

juanmardefago: I'm referring to `fetchSubgraphDeploymentManifest`, which would already fetch the manifest and store it as a String representation of the file in the `SubgraphDeployment` entity btw

juanmardefago:
Another thing I forgot to mention (and this one is on me) is that in our workflow we first merge to the branch `mainnet-staging` and then move it to `master` once it's synced and tested in staging environments.

I should probably update the `README.md` as well as `CONTRIBUTING.md` files, since there's probably no mention of this anywhere

src/mappings/gns.ts
mnlesane: It was a bit of a design relic from the original subgraph, but seeing as the manifest is only retrieved once per SubgraphDeployment entity, I replaced the IPFS retrieval with processing SubgraphDeployment.manifest and moved the call to createOrLoadSubgraphDeployment.  Let me know if i I should move the call into fetchSubgraphDeploymentManifest instead.

juanmardefago:
@mnlesane sorry for the long delay, I've been caught in a ton of stuff and missed these updates.

Overall it looks a lot cleaner now and I really like the approach!

Only thing missing is to double check that it's working as expected (I'd love to have tests in the subgraph soon, but for now at least double checking it on a synced version would be just fine)

mnlesane:
@juanmardefago All changes have been verified via a [fork](https://thegraph.com/hosted-service/subgraph/mnlesane/graph-network-mainnet) that was deployed to the hosted service.  I wrote a simple test a while back for standardizeAddresses(), which I've added assertions to and checked into this branch.  If unit tests should also be created for the parsing logic, please let me know.

juanmardefago:
I mentioned I was going to merge this on #192, but it seems that the link you sent has a failed subgraph there. Just double checking, is the failed subgraph running the same code as this branch?

juanmardefago:
Closing since this is quite outdated, and we'll be working on supporting this feature properly in the future

PR #185, opened 2022-04-27
Slightly more robust manifest parsing

juanmardefago:
This change tries to improve the parsing code to make sure it doesn't fail if it can't find the basic expected format for a "valid" manifest. (I'm not entirely sure there's an official spec that would unambiguously define what is a "valid" manifest, but from what I could see being generated and the minimum requirements you need for it to be indexable, this should be as close as we have to a spec right now)


IainM22 (approved):
Fantastic

PR #192, opened 2022-08-06
Assign creatorAddress in handleSubgraphPublishedV2

mnlesane:
The creatorAddress field has not been assigned for subgraphs created after 2022-02-23, as the graphAccount param is no longer associated with the SubgraphPublished event for v2.  This PR uses event.transaction.from instead.

Demonstration:
https://thegraph.com/hosted-service/subgraph/mnlesane/graph-network-mainnet
```{
  subgraphs(orderBy:createdAt, orderDirection:desc
  ) {
    id displayName creatorAddress
  }
}```

### DIFF
--- a/src/mappings/gns.ts
+++ b/src/mappings/gns.ts
@@ -760,6 +760,7 @@ export function handleSubgraphPublishedV2(event: SubgraphPublished1): void {
   subgraph.reserveRatio = event.params.reserveRatio.toI32()
   subgraph.migrated = true
   subgraph.initializing = true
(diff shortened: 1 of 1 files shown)


juanmardefago:
Hey @mnlesane!
Looks good! I'm going to merge it to mainnet-staging now that we got the workflows back to a state where it's only going to be deploying to staing on pushes there.

I'm also gonna merge #182 there, and check if everything works as expected afterwards, before moving it to production!

Thanks a ton for the contributions and sorry for the long delays!

PR #194, opened 2022-08-30
added github worlfow templates to remove duplicated code

rotarur:


.github/workflows/template-deploy.yaml
```diff
@@ -0,0 +1,73 @@
...
+  template-deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set staging environment variables
+        if: ${{ inputs.ENVIRONMENT == 'staging' }}
+        run: |
+          echo "ENV_SUFFIX=-staging" >> $GITHUB_ENV
+
+      - name: Set goerli production environment variables
```
Theodus, line 26:
```suggestion
      - name: Set mainnet staging environment variables
```

.github/workflows/template-deploy.yaml
```diff
@@ -0,0 +1,73 @@
+name: Production Subgraph (Mainnet)
```
Theodus, line 1: I don't think production is limited to mainnet. In the context of this repo, testnets like goerli, rinkeby, etc. are production. This is also used for mainnet staging, so I think this name could lead to confusion.

Theodus (changes requested):
I think there should be more consistency around naming, especially as we will be adding support for Arbitrum soon. I suggest using the following environment names:
- mainnet
- testnet-goerli
- testnet-rinkeby (obsolete?)
- mainnet-staging (or simply staging, if we don't expect multiple staging environments)

I'd like to get an opinion from @juanmardefago on this, since he has more insight into the testing/release flow for this codebase.

.github/workflows/template-deploy.yaml
rotarur: which one do you recommend? should I use `Production Mainnet` ? it was like this before this PR

.github/workflows/template-deploy.yaml
rotarur: thanks, fixed

.github/workflows/template-deploy.yaml
Theodus: How about this?
```suggestion
name: Deploy Subgraph (${{ inputs.ENVIRONMENT }})
```

Theodus (approved):
LGTM

.github/workflows/template-deploy.yaml
```diff
@@ -0,0 +1,73 @@
...
+        run: yarn install
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NODE_AUTH_TOKEN }}
+
+      # Run scripts
+      - name: Prepare IPFS enabled files
+        run: ./node_modules/.bin/mustache ./config/ipfs.json ./src/mappings/metadataHelpers.template.ts > ./src/mappings/metadataHelpers.ts
+
+      - name: Prepare addresses ${{ inputs.ENVIRONMENT }}
+        run: ./node_modules/.bin/ts-node config/mainnetAddressScript.ts && ./node_modules/.bin/mustache ./config/generatedAddresses.json ./config/addresses.template.ts > ./config/addresses.ts
```
juanmardefago, line 67: @Theodus @rotarur each of the environments uses a different config file in this step, in this case, the template always uses the mainnet file, and thus, will always deploy mainnet code, even if you change the endpoint where it's going to deploy to

.github/workflows/template-deploy.yaml
rotarur: this doesn't work like that, the variables are only known later, so the name can't be dynamically created

.github/workflows/template-deploy.yaml
rotarur: @juanmardefago thanks, please check the lastest commit

.github/workflows/template-deploy.yaml
Theodus: Ah sorry. Just `Deploy Subgraph` then

rotarur: @Theodus please check my latest commit

Theodus (approved):
LGTM

juanmardefago:
Looks good now!

juanmardefago (approved):

PR #204, opened 2023-01-13
feat: refactor usage of block number for L2 deployments

juanmardefago:
In order to line up with what the contracts do (use L1 blocks rather than L2 blocks for all calculations) we have to refactor the subgraph in order to be able to have those L1 blocks rather than the chain native L2 blocks for both Arbitrum and Arbitrum Goerli, without impacting performance on the L1 deployments.


src/mappings/curation.ts
```diff
@@ -66,7 +67,7 @@ export function handleSignalled(event: Signalled): void {
 
   signal.signal = signal.signal.plus(event.params.signal)
   signal.lastUpdatedAt = event.block.timestamp.toI32()
```
tmigone, line 69: I would also change all the `lastUpdatedAt` to use `graphNetwork.currentL1BlockNumber` for L2 subgraphs. The block number reported by the `EpochManager` on L2 uses `block.number` which according to arbitrum's [docs](https://developer.arbitrum.io/time#ethereum-block-numbers-within-arbitrum) _return a value close to (but not necessarily exactly) the L1 block number_. The difference is probably subtle but it might cause some debugging headaches down the line if we don't keep track of it.

tmigone (changes requested):
Looks good, just left one comment about `lastUpdatedAt` which we should consider.

src/mappings/curation.ts
juanmardefago: Good catch, but unfortunately for this, we don't use the block number, but rather the timestamp, and we don't have the exact timestamp for the L1 block available in this context (nor we can get it with EpochManager).

I think given that it's a timestamp, it shouldn't cause too much troubles, unless we try to link the timestamp to an L1 block for some debugging reason. (even so, this field is mostly just aesthetic, not really used anywhere for anything more than displaying it as extra information)

tmigone: Ah right we can't make RPC calls inside a subgraph lol. I guess we could add a new function to the EpochManager that returns the `block.timestamp` same as we have with the `block.number`, but if it's as you say then maybe it's not worth the trouble of getting this through the council, deployed, etc.

tmigone: I'll add a very low prio task to include a `blockTimestamp()` function to the epoch manager, we never know where else this might come in handy. I don't think I'll push to get it deployed, accepted into mainnet, etc anytime soon, but eventually i'll get there and we can update this subgraph.

tmigone (approved):

PR #205, opened 2023-01-18
feat: add bridge deposits and withdrawals to the network subgraph

tmigone:
This PR adds the ability to track bridge withdrawals across L1 and L2 on the network subgraph using "primary key" type unique values to match on both subgraphs.

**L2 to L1**
A new entity `BridgeWithdrawalTransaction` is created with a unique value, `transactionIndex`, that allows matching an L2 transaction initiating a withdrawal with its L1 counterpart finalizing it.

Note that while we use the same entity to store the withdrawal data on L1 and L2, the way we get that data is different depending on the chain we are in:
- On L2, the `L2GraphTokenGateway` emits an event with the actual `transactionIndex` value.
- On L1, the `L1GraphTokenGateway` emits an event but the key value `transactionIndex` is missing. Fortunately this can be retrieved from a different event that's triggered on the same transaction by the arbitrum `Outbox` contract. To get this data we use the [transaction receipts feature](https://thegraph.com/docs/en/developing/creating-a-subgraph/#transaction-receipts-in-event-handlers) which forces us to update some dependencies and the subgraph's `specVersion` (hope this doesn't break everything 🤞 ).

**L1 to L2**
A new entity `BridgeDepositTransaction` is created with a unique value, `retryableTicketId`, that allows matching an L1 transaction initiating a deposit with its L2 counterpart finalizing it.

Again, we use the same entity on both layers but the way the retryable ticket id is computed differs:
- On L2, there are two events we need to watch. First we index the `ArbRetryTx` precompile and store all retryable ticket creations alongside their redeem attempt. We then index the `DepositFinalized` events from the `L2GraphTokenGateway` and match the tx hash with the retryable redeem attempt to figure out the retryable ticket id.
- On L1, the `L1GraphTokenGateway` emits several events with data that can be used to compute the retryable ticket id (a lot more easier said than done). This requires using the transaction receipts feature described above.

### Considerations
On both bridge operations (L1 to L2, L2 to L1) we require data from multiple events on a tx to compute the value that acts as unique key (`transactionIndex` or `retryableTicketId`). The way this is implemented means that under certain conditions the calculation will fail and return `null`, namely if multiple deposits/withdrawals are performed on the same tx (for example using a multicall to make several bridge txs). This is not to be expected for the regular user bridging assets but it's something we could monitor and eventually decide upon improving the subgraph heuristics. 

### Sync time

Added a `startBlock` to all new events. Almost all synchronise in a couple of seconds. 
Worth to note that indexing the `ArbRetryTx` precompile takes a bit longer since it's indexing stuff probably in most blocks (takes about 30 minutes in arbitrum one). 

### Changes
- Schema changes
   - Add `BridgeWithdrawalTransaction` and `BridgeDepositTransaction` entities.
   - Implement `totalGRTWithdrawnConfirmed` and `totalGRTDepositedConfirmed` in `GraphNetwork` entity.
   - Add `bridgeWithdrawalTransactions`  and `bridgeDepositTransactions` to `GraphAccount` entity.
- Subgraph changes due to using transaction receipts feature
   - Upgrade `graph-cli` to `^0.37.5` and `graph-ts` to `0^.29.1`
   - Upgrade `specVersion` to `0.0.5` and `apiVersion` to `0.0.7`
- Moved all helpers into a `helpers` directory


subgraph.template.yaml
```diff
@@ -472,25 +478,28 @@ dataSources:
...
+      language: wasm/assemblyscript
+      file: ./src/mappings/arbRetryableTx.ts
+      entities:
+        - GraphNetwork
+      abis:
+        - name: ArbRetryableTx
+          file: ./abis/ArbRetryableTx.json
+      eventHandlers:
+        - event: RedeemScheduled(indexed bytes32,indexed bytes32,indexed uint64,uint64,address,uint256,uint256)
+          handler: handleRedeemScheduled
```
juanmardefago, line 502: Is this address correct? and if so, should it also be hardcoded for all L1s?

config/arbitrumGoerliAddressScript.ts
```diff
@@ -35,7 +35,7 @@ const main = (): void => {
   try {
     let output = JSON.parse(mustache.render(JSON.stringify(addresses), renameAddresses))
     output.blockNumber = '1023264' // Protocol deployment
-    output.bridgeBlockNumber = '1023272' // Bridge deployment block
+    output.bridgeBlockNumber = '1023318' // Bridge deployment block on L2
```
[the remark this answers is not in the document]
tmigone: I think you are right, I don't remember why I picked `1023318` but looking at the `L2GraphTokenGateway` here https://goerli.arbiscan.io/address/0x18C924BD5E8b83b47EFaDD632b7178E2Fd36073D, `1023272` is the block where the contract was created so I'll revert the value.

src/mappings/helpers/bridge.ts
```diff
@@ -0,0 +1,159 @@
...
+
+// Gets transactionIndex
+// Returns null if the withdrawal call is made under the following conditions:
+// - Call is made from another contract (i.e: a multicall call) and
+// - Call contains multiple 'OutBoxTransactionExecuted' events (i.e: multiple withdrawals on same multicall call)
+export function getTransactionIndex(event: ethereum.Event): BigInt | null {
+  return (
+    getTransactionIndexFromCalldata(event) || getTransactionIndexFromLogs(event)
+  );
+}
```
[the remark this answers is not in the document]
tmigone: In my limited experience testing it, it did work 😓

subgraph.template.yaml
tmigone: Yes it's correct! It's an Arbitrum pre-compile, meaning it's a "contract" baked in the arbitrum node, always has the same address regardless of the environment.

tmigone: If you are curious here you can read more about them: https://developer.arbitrum.io/arbos/common-precompiles

juanmardefago (approved):

PR #208, opened 2023-02-02
feat: 1:1 registry, filled testnet address

hopeyen:
Include gossip registry within the network subgraph
- Handler
- ABI file
- Contract addresses


juanmardefago:
Closing since we moved to a solution outside the network subgraph

PR #216, opened 2023-03-02
fix: off-by-one error on new environments

juanmardefago:
Due to [this change](https://github.com/graphprotocol/contracts/commit/b352e49c4aabec381f412b0f9b5306d0638ac385), new environments are no longer initialized on epoch 0. To properly reflect the change (but also support older enviornments, i.e. mainnet) the change uses the defined epoch from the event triggered in the constructor as a separate case to properly initialize the first epoch.
Fix can be checked working on [mainnet](https://thegraph.com/hosted-service/subgraph/juanmardefago/dev-subgraph) and [goerli](https://thegraph.com/hosted-service/subgraph/juanmardefago/dev-subgraph2) dev subgraphs, comparing it to the EpochManager currentEpoch, and to prior versions of [mainnet](https://thegraph.com/hosted-service/subgraph/graphprotocol/graph-network-mainnet) and [goerli](https://thegraph.com/hosted-service/subgraph/graphprotocol/graph-network-goerli) (mainnet was ok, goerli was off-by-one as it was a newer environment)


tmigone (approved):
LGTM!

PR #217, opened 2023-03-23
feat: add support for post-GIP-0037 issuance

pcarranzav:


tmigone:
We still need to update the `@graphprotocol/contracts` version to v3 once it's released

src/mappings/helpers/helpers.ts
```diff
@@ -503,6 +503,7 @@ export function createOrLoadGraphNetwork(
     graphNetwork.totalIndexingDelegatorRewards = BigInt.fromI32(0)
 
     graphNetwork.networkGRTIssuance = BigInt.fromI32(0)
+    graphNetwork.networkGRTIssuancePerBlock = BigInt.fromI32(0)
```
juanmardefago, line 506: We should also initialize `totalGRTMintedFromL2` here.

juanmardefago (approved):

PR #218, opened 2023-04-05
fix: issues with L1GraphTokenGateway json on 2.3.0 and rewards manager stitched

juanmardefago:
Fixes deployment issues from #217

### DIFF
--- a/abis/L1GraphTokenGateway.json
+++ b/abis/L1GraphTokenGateway.json
@@ -1,0 +1,902 @@
+[
+    {
+      "anonymous": false,
+      "inputs": [
+        {
+          "indexed": false,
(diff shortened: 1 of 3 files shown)


juanmardefago:
@tmigone @pcarranzav unfortunately #217 had 2 issues we could only see after deploying, one regarding the L1GraphTokenGateway.json ABI found in contracts 2.3.0, which for some reason doesn't contain the TokensMintedFromL2 event, getting the latest ABI from 4.0.0-scratch2r1 fixed the issue, but we might want to revert the change once the latest stable NPM package has that event in it.
The other issue was an import issue, the newly stitched RewardsManager ABI wasn't being referenced, but rather the old one was.

PR #219, opened 2023-04-06
L2 subgraph migration helpers

juanmardefago:
* Added support for new migration flags on both L1 and L2
 * On L1 only the ones linked to `startedMigrationToL2` will be available, as we don't know when it actually finishes. The rest will be available but will be null.
  * On L1 both `startedMigrationToL2` and `migratedToL2` will be available, with `startedMigrationToL2` being set to true when the migration is detected to start in L2, and `migratedToL2` when it's finished and the subgraph is published.
  * `idOnL2` and `idOnL1` will be available on both L1 and L2 but ONLY when there's a migration event triggered (i.e., if the L1 subgraph hasn't been migrated to L2, both fields will be null, the same for any L2 subgraph that hasn't been created through a migration)
* Renamed `Curation` to `L2Curation` on the addresses helpers for L2, since the npm package also did this.
* Added `Controller` ABI access to TokenLockManager data source since for goerli 4.0.0-scratch2r1 there's actually a vesting contract manager that has events before other contracts, which lead to weird interactions that hadn't happened in other environments


juanmardefago:
Deployed L1 (Goerli) to this endpoint: https://thegraph.com/hosted-service/subgraph/juanmardefago/dev-subgraph?version=pending
Deployed L2 (Arbitrum-Goerli) to this endpoint: https://thegraph.com/hosted-service/subgraph/juanmardefago/dev-subgraph2?version=pending

juanmardefago:
It seems goerli is having some issues getting the governor address, maybe the vesting contracts manager was initialized and running before other core contracts?

PR #221, opened 2023-04-11
Added addresses.ens Check

Data-Nexus:
Added a check for addresses.ens to `verifyNameOwnership()`. Arbitrum does not have an address and when attempting to get it, the subgraph errors.


juanmardefago:
Closing since this was solved, but thanks for contributions as always @Data-Nexus!

PR #230, opened 2023-06-20
Fix ownStakeRatio calculation with locked tokens

zekchan:


juanmardefago:
Thanks for the contribution, and sorry I missed it!

I'll change the target branch to `mainnet-staging` so we can add it to the release pipeline appropiately :)

PR #236, opened 2023-08-02
Support exponential rebates upgrade

tmigone:
The following is a list of changes to the `Staking` and `RewardsManager` contract introduced by the exponential rebates upgrade and how they affect the network subgraph. All deprecated variables and entities should still be updated and valid up to the block where the exponential rebates upgrade was performed:

- Replace Cob-Douglass rebates with exponential rebates
   - [x] Add `lambdaNumerator` and `lambdaDenominator` to staking contract storage. For `alphaNumerator` and `alphaDenominator` we will re-use the storage variables currently in use for the Cobb-Douglas formula.
      - New variables in `GraphNetwork` entity: `rebateAlpha`, `rebateLambda`
      - Deprecated in `GraphNetwork` entity: `rebateRatio`
- Update `collect()` function to use exponential rebates
   - [x] Remove `claim()` function. Combine it with `collect()` in a single operation for better indexer UX. Replace `AllocationCollected` with `RebatesCollected` event
      - Added `RebatesCollected` event and handler combining logic from both previous events `AllocationCollected` and `RebateClaimed`
   - [x] Add `distributedRebates` to the `Allocation` struct to keep track of rebate distribution across multiple voucher collections.
      - Added `distributedRebates` variable to `Allocation` entity.
- Remove rebate pools 🥳 
   - [x] Deprecate `rebates` storage variable. Update `closeAllocation()` function to remove rebate pools logic.
      - Deprecated `Pool` entity and `poolClaimedIn` variable in `Allocation` entity.
      - Add a second `AllocationClosed` event and handler to match both event signatures (before and after upgrade).
   - [x] Deprecate `channelDisputeEpochs` storage variable and remove associated setters/getters
      - Deprecated in `GraphNetwork` entity: `channelDisputeEpochs`
   - [x] Deprecate `effectiveAllocation` in `Allocation` struct
      - Deprecated `effectiveAllocation` variable in `Allocation` entity.
   - [x] Remove `Claimed` and `Finalized` states from `Allocation` life cycle, affects `getAllocationState()`
      - Deprecated `Claimed` and `Finalized` states from `AllocationStatus` enum, which means `Allocation` entities variable `status` can no longer be in those two states. For consistency, this includes all history, even allocations claimed prior to the upgrade.

### DIFF
--- a/abis/IStakingStitched.json
+++ b/abis/IStakingStitched.json
@@ -1,0 +1,1994 @@
+[
+  {
+    "inputs": [],
+    "name": "channelDisputeEpochs",
+    "outputs": [{ "internalType": "uint32", "name": "", "type": "uint32" }],
+    "stateMutability": "view",
+    "type": "function"
+  },
+  {
+    "anonymous": false,
+    "inputs": [
+      { "indexed": true, "internalType": "address", "name": "indexer", "type": "address" },
+      {
+        "indexed": true,
+        "internalType": "bytes32",
+        "name": "subgraphDeploymentID",
+        "type": "bytes32"
+      },
+      { "indexed": false, "internalType": "uint256", "name": "epoch", "type": "uint256" },
+      { "indexed": false, "internalType": "uint256", "name": "tokens", "type": "uint256" },
+      { "indexed": true, "internalType": "address", "name": "allocationID", "type": "address" },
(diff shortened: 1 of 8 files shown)


pcarranzav (approved):
Approving without review to fix production issues

PR #244, opened 2023-09-01
File data sources refactor

juanmardefago:


schema.graphql
```diff
@@ -390,7 +396,7 @@ type Subgraph @entity {
   initializing: Boolean!
   "Version of the entity. Subgraph entities are changing the way their ID is generated when the new GNS v2 rolls out so we need to differnetiate them"
   entityVersion: Int!
-  "Subgraph entities might have 'duplicate' entities so that old IDs can be persisted and still queried properly. If that happens we have to link both duplicate entities to be able to update them properly."
+  "[DEPRECATED] Used for duplicate entities to enable old IDs from before the subgraph NFT update"
```
hasparus, line 399: @juanmardefago What do you think about using `@deprecated` directive?

It would allow us to migrate from deprecated fields in the frontend with the help of The Guild's ESLint plugin.

https://the-guild.dev/graphql/eslint/rules/no-deprecated

juanmardefago: Wasn't even aware this existed, although if we wanted to go that route we'd might aswell add those directives to all past deprecations aswell

juanmardefago: @hasparus I'm gonna leave that as a future refactor so that it's self contained, and after that we can use that deprecation procedure by default

PR #253, opened 2023-11-23
fix: added arbitrum sepolia to get retryable tickets

Maikol:
Added Arbitrum Sepolia network for getting retryable tickets. Included a critical log so next time we add a new chain we don't miss this step.

### DIFF
--- a/src/mappings/helpers/bridge.ts
+++ b/src/mappings/helpers/bridge.ts
@@ -59,7 +59,7 @@ export function getRetryableTicketId(
   } else if (addresses.network === "sepolia") {
     l2ChainIdHex = "0x066EEE"; // 0x066EEE = 421614 (Arbitrum Sepolia)
   } else {
-    log.critical('Unsupported network: {}', [addresses.network]);
+    log.warning('Unsupported network: {}', [addresses.network]);
     return null;
   }
 
--- a/src/mappings/helpers/bridge.ts
+++ b/src/mappings/helpers/bridge.ts
@@ -4,6 +4,7 @@ import {
   ethereum,
   crypto,
   ByteArray,
+  log,
 } from "@graphprotocol/graph-ts";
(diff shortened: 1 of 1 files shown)


tmigone (approved):

src/mappings/helpers/bridge.ts
```diff
@@ -50,9 +51,18 @@ export function getRetryableTicketId(
...
-  let l2ChainIdHex = addresses.network === "mainnet" ? "0xA4B1" : "0x066EED";
+  let l2ChainIdHex = "";
+  if (addresses.network === "mainnet") {
+    l2ChainIdHex = "0xA4B1"; // 0xA4B1 = 42161 (Arbitrum One)
+  } else if (addresses.network === "goerli") {
+    l2ChainIdHex = "0x066EED"; // 0x066EED = 421613 (Arbitrum Goerli)
+  } else if (addresses.network === "sepolia") {
+    l2ChainIdHex = "0x066EEE"; // 0x066EEE = 421614 (Arbitrum Sepolia)
+  } else {
+    log.critical('Unsupported network: {}', [addresses.network]);
```
juanmardefago, line 62: This would break all deployments except mainnet, goerli and sepolia, since `log.critical` causes the subgraph to crash. If we want to give a warning or error message, please change it to `log.error` or `log.warning`

juanmardefago (changes requested):

src/mappings/helpers/bridge.ts
Maikol: I see, my intention was to be loud so next time we wouldn't forget to update this but we can't do that since we do deploy this to other networks. I'll update to a warning.

juanmardefago: Error would probably be also a decent choice if you want to make it extra obvious that something isn't expected, but given the context warning makes sense too

juanmardefago (approved):

PR #257, opened 2024-01-29
fix: track total delegated tokens changes when rewards or fees are collected

pcarranzav:


juanmardefago (approved):

juanmardefago:
Indexer side of the totals should be already taken care of, as restaking for indexers is done through `_sendRewards` and if restaked, it triggers a `StakeDeposited` event which takes care of the network totals

PR #261, opened 2024-01-31
fix: workaround query type issues if entity is named Metadata

juanmardefago:
Currently syncing [here](https://thegraph.com/hosted-service/subgraph/juanmardefago/dev-subgraph?version=pending) for Arbitrum-one for testing purposes


juanmardefago:
This does introduce breaking changes since it's a schema change.

PR #280, opened 2025-02-11
[Horizon] ProvisionManagement implementation

juanmardefago:
Uses work on PR #279 as base


src/mappings/horizonStaking.ts
```diff
@@ -72,3 +83,135 @@ export function handleHorizonStakeWithdrawn(event: HorizonStakeWithdrawn): void
...
+    // To Do
+}
+
+export function handleProvisionSlashed(event: ProvisionSlashed): void {
+    let graphNetwork = createOrLoadGraphNetwork(event.block.number, event.address)
+    let dataService = createOrLoadDataService(event.params.verifier)
+    let indexer = Indexer.load(event.params.serviceProvider.toHexString())!
+    let provision = createOrLoadProvision(event.params.serviceProvider, event.params.verifier, event.block.timestamp)
+
+    // Due to thawing tokens potentially getting cancelled, we will need to figure the thawing situation
```
tmigone, line 181: Would it help if we emitted extra events whenever the thawing cancelling happens? 

We could do `event ThawingPoolReset(string type, uint256 newNonce)`

src/mappings/horizonStaking.ts
```diff
@@ -72,3 +83,135 @@ export function handleHorizonStakeWithdrawn(event: HorizonStakeWithdrawn): void
...
+    indexer.save()
+
+    dataService.totalTokensProvisioned = dataService.totalTokensProvisioned.minus(event.params.tokens)
+    dataService.save()
+
+    graphNetwork.totalTokensProvisioned = graphNetwork.totalTokensProvisioned.minus(event.params.tokens)
+    graphNetwork.totalTokensStaked = graphNetwork.totalTokensStaked.minus(event.params.tokens)
+    graphNetwork.save()
+
+    provision.tokensProvisioned = provision.tokensProvisioned.minus(event.params.tokens)
```
tmigone, line 193: we also need to update `provision.tokensThawing`

schema.graphql
```diff
@@ -838,6 +846,44 @@ type Provision @entity {
...
+  maxVerifierCutPending: BigInt!
+
+  "Time, in seconds, tokens must thaw before being withdrawn"
+  thawingPeriod: BigInt!
+
+  "Pending value for thawingPeriod. Verifier needs to accept it before it becomes active"
+  thawingPeriodPending: BigInt!
+}
+
+type ThawRequest @entity {
```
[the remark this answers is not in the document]
tmigone: Ok i see that we are not tracking the list on the subgraph which is probably a good decision at this point, so I take back my previous comment :D

schema.graphql
```diff
@@ -838,6 +846,44 @@ type Provision @entity {
...
+
+  "Shares representing the thawing tokens"
+  shares: BigInt!
+
+  "Tokens thawed"
+  tokens: BigInt!
+
+  thawingUntil: BigInt!
+
+  valid: Boolean
```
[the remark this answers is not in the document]
tmigone: Maybe also a `fulfilled` boolean to give more meaning to the `fulfilledAsValid` property? Not sure it's needed i'll leave it up to you.

src/mappings/horizonStaking.ts
```diff
@@ -72,3 +83,135 @@ export function handleHorizonStakeWithdrawn(event: HorizonStakeWithdrawn): void
...
+    indexer.provisionedTokens = indexer.provisionedTokens.minus(event.params.tokens)
+    indexer.save()
+
+    dataService.totalTokensProvisioned = dataService.totalTokensProvisioned.minus(event.params.tokens)
+    dataService.save()
+
+    graphNetwork.totalTokensProvisioned = graphNetwork.totalTokensProvisioned.minus(event.params.tokens)
+    graphNetwork.save()
+
+    provision.tokensProvisioned = provision.tokensProvisioned.minus(event.params.tokens)
```
tmigone, line 163: missing `tokensThawing` update here

tmigone (commented):
🚀

schema.graphql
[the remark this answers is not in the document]
juanmardefago: Yeah makes sense. Will update

src/mappings/horizonStaking.ts
juanmardefago: IIRC the main issue was figuring out both if the thaw requests got cancelled (which I seem to recall we discussed it and it was under a very specific condition that all tokens were gone from the slash) but also which thaw requests have to be canceled, since we don't delete thaw request entities, and loading all of them up just to check one by one which one was fulfilled already and which ones were still active doesn't make a lot of sense.

src/mappings/horizonStaking.ts
juanmardefago: This one is actually a little bit weird, will discuss it on monday if possible

PR #281, opened 2025-02-13
[Horizon] ProvisionConfiguration implementation

juanmardefago:


schema.graphql
```diff
@@ -837,6 +839,21 @@ type IndexerQueryFeePaymentAggregation @entity {
   paymentSource: PaymentSource!
 }
 
+type Operator @entity {
+  "Joint address of all 3 parties involved (Operator, Indexer, DataService)"
+  id: ID!
```
tmigone, line 844: Would there be any advantages if we had one operator per address then tracked which (indexer, dataservice) combinations they have or dont have access to? From an outsider's perspective it seems more intuitive for querying the data, but im not sure if it's a better model for the subgraph code.

schema.graphql
```diff
@@ -273,6 +273,8 @@ type GraphAccount @entity {
   operatorOf: [GraphAccount!]! @derivedFrom(field: "operators")
   "Operators of this Graph Accounts"
   operators: [GraphAccount!]!
+  "Operators of this Graph Accounts"
```
tmigone, line 276:
```suggestion
  "Horizon operators of this Graph Accounts"
```

src/mappings/horizonStaking.ts
```diff
@@ -165,11 +165,67 @@ export function handleTokensDeprovisioned(event: TokensDeprovisioned): void {
...
+    provision.thawingPeriodPending = event.params.thawingPeriod
+    provision.maxVerifierCutPending = event.params.maxVerifierCut
+    provision.save()
+}
+
+export function handleOperatorSet(event: OperatorSet): void {
+    let indexerGraphAccount = createOrLoadGraphAccount(event.params.serviceProvider, event.block.timestamp)
+    let operator = createOrLoadOperator(event.params.operator, event.params.verifier, event.params.serviceProvider)
+    let operators = indexerGraphAccount.operators
+    // Will have to handle legacy operators list, and horizon provisionedOperators list for extra context
```
tmigone, line 185: Why cant we deprecate and remove the legacy operators list? Just process event history and add to the new operator list?

tmigone (commented):
looks great!

schema.graphql
```diff
@@ -837,6 +839,21 @@ type IndexerQueryFeePaymentAggregation @entity {
   paymentSource: PaymentSource!
 }
 
+type Operator @entity {
```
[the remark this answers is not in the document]
juanmardefago: Yeah, I just wanted to keep the name length small since those would always be horizon operators, but we could change it for clarity purpose.

schema.graphql
```diff
@@ -837,6 +839,21 @@ type IndexerQueryFeePaymentAggregation @entity {
...
+  "Joint address of all 3 parties involved (Operator, Indexer, DataService)"
+  id: ID!
+
+  operatorOf: [GraphAccount!]! @derivedFrom(field: "provisionedOperators")
```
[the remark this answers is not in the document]
juanmardefago: I think we were already using `operatorOf` before, so we just stuck with the same naming for cohesion.

schema.graphql
```diff
@@ -864,6 +881,21 @@ type Provision @entity {
 
   "Pending value for thawingPeriod. Verifier needs to accept it before it becomes active"
   thawingPeriodPending: BigInt!
+
+  # Cuts for subgraph service. Percentage of rewards that the delegation pool receives
```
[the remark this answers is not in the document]
juanmardefago: Yeah, I mentioned it as "for the subgraph service" as they were tailored for it, but you are right that the aren't particularly exclusive

schema.graphql
```diff
@@ -864,6 +881,21 @@ type Provision @entity {
...
+  queryFeeCut: BigInt!
+
+  indexingFeeCut: BigInt!
+
+  indexingRewardsCut: BigInt!
+
+  # Might want to add polymorphic handling of different fee cuts through a generic hashmap in the future
```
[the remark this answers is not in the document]
juanmardefago: Fancy stuff not yet implemented 😅

schema.graphql
```diff
@@ -864,6 +881,21 @@ type Provision @entity {
...
+}
+
+enum PaymentTypes {
```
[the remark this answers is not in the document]
juanmardefago: Can't remember why I added it, maybe I just did at the beginning and ended up not using it 😅. Could be used for the polymorphic fancy implementation in the future though, albeit it would have to be constantly updated to make sense, so it might not be worth keeping

schema.graphql
juanmardefago: Not sure I follow 🤔 

I think it's not super complex as it is (it makes sense that the ID is the joint ID of all 3 parties involved, so, the operator address, the indexer and the data service, or rather, the operator address and the provision "ID")

src/mappings/horizonStaking.ts
juanmardefago: I think it was for backwards compatibility and also the new operator list is for operator <-> Provision relationship rather than Operator <-> Indexer relationship, but I don't really remember now, will check soon

juanmardefago: Yup, essentially the new HorizonOperator needs to know it's provision. We could make it optional and add everything to the same operator list, but... that unnecessarily breaks compatibility for the old stuff.

We could do it so that everything is in the same place, but given we will have a Stage 2 subgraph that essentially erases all pre-Horizon data, having a clear cut as to how to support each side, while still making it compatible with the legacy way (but only for legacy data) seemed to make the most sense

PR #282, opened 2025-02-18
[Horizon] DelegationPoolManagement implementation

juanmardefago:


schema.graphql
```diff
@@ -890,6 +890,14 @@ type Provision @entity {
   indexingRewardsCut: BigInt!
 
   # Might want to add polymorphic handling of different fee cuts through a generic hashmap in the future
+
+  # Delegation
+  "Tokens delegated to the provision"
+  delegatedTokens: BigInt!
+  "Total shares of the delegator pool"
+  delegatorShares: BigInt!
+  "Exchange rate of of tokens received for each share"
```
tmigone, line 899:
```suggestion
  "Exchange rate of tokens received for each share"
```

src/mappings/horizonStaking.ts
```diff
@@ -270,4 +262,14 @@ export function handleThawRequestFulfilled(event: ThawRequestFulfilled): void {
     request.tokens = event.params.tokens
     request.valid = event.params.valid
     request.save()
+}
+
+export function handleTokensToDelegationPoolAdded(event: TokensToDelegationPoolAdded): void {
+    let indexer = Indexer.load(event.params.serviceProvider.toHexString())!
+    let provision = createOrLoadProvision(event.params.serviceProvider, event.params.verifier, event.block.timestamp)
+    provision.delegatedTokens = provision.delegatedTokens.plus(event.params.tokens)
```
tmigone, line 270: We likely also want to update `provision.delegationExchangeRate` here no? We are not issuing new shares but the tokens backing those shares change.

src/mappings/horizonStaking.ts
```diff
@@ -270,4 +262,14 @@ export function handleThawRequestFulfilled(event: ThawRequestFulfilled): void {
...
+    provision.save()
+
+    indexer.delegatedTokens = indexer.delegatedTokens.plus(event.params.tokens) // this only serves as a general tracker, but the real deal is per provision
```
tmigone, line 273: similar to previous comment, there is also `indexer.delegationExchangeRate`

src/mappings/horizonStaking.ts
juanmardefago: I'm not sure but I think I updated this when I worked on the delegation code (as I worked on this a long time before getting to the full delegation support PR), will double check and fix if missing

PR #285, opened 2025-02-24
[Horizon] AllocationManagement implementation

juanmardefago:


src/mappings/subgraphService.ts
```diff
@@ -20,3 +22,148 @@ export function handleRewardsDestinationSet(event: RewardsDestinationSet): void
...
+    allocation.indexingRewards = BigInt.fromI32(0)
+    allocation.indexingIndexerRewards = BigInt.fromI32(0)
+    allocation.indexingDelegatorRewards = BigInt.fromI32(0)
+    allocation.delegationFees = BigInt.fromI32(0)
+    allocation.status = 'Active'
+    allocation.totalReturn = BigDecimal.fromString('0')
+    allocation.annualizedReturn = BigDecimal.fromString('0')
+    allocation.createdAt = event.block.timestamp.toI32()
+    allocation.indexingRewardCutAtStart = provision.indexingRewardsCut.toI32()
+    allocation.indexingRewardEffectiveCutAtStart = provision.indexingRewardsCut.toBigDecimal()
```
tmigone, line 85: this is incorrect no? we need to calculate with helper function

src/mappings/subgraphService.ts
```diff
@@ -20,3 +22,148 @@ export function handleRewardsDestinationSet(event: RewardsDestinationSet): void
...
+    allocation.queryFeeCutAtStart = provision.queryFeeCut.toI32()
+    allocation.queryFeeEffectiveCutAtStart = provision.queryFeeCut.toBigDecimal()
```
tmigone, line 87: same, i think it's incorrect and we need to calculate

schema.graphql
```diff
@@ -934,17 +943,21 @@ type ThawRequest @entity {
 }
 
 """
-A state channel Allocation representing a single Indexer-SubgraphDeployment stake
+A state channel Allocation representing a single Indexer/Provision-SubgraphDeployment stake
 """
 type Allocation @entity {
```
tmigone, line 948: we need to add a boolean `isLegacy` to signal wether the allocation was originally created on the `Staking.sol` contract before Horizon, or on the `SubgraphService.sol` contract post horizon

schema.graphql
```diff
@@ -987,10 +1000,12 @@ type Allocation @entity {
   createdAt: Int!
   "Timestamp this allocation was closed at"
   closedAt: Int
-  "POI submitted with a closed allocation"
+  "POI submitted when closing allocations in the legacy protocol. Also reflects the latest POI submitted for a Horizon payment collection"
   poi: Bytes
+  "POIs submitted when collecting payments in Horizon"
+  pois: [PoiSubmission!]! @derivedFrom(field: "allocation")
```
tmigone, line 1006: maybe add a `poiCount`?

schema.graphql
tmigone: Another nice addition would be a `forceClosed` boolean. An allocation is force closed when the msg.sender is not the indexer nor an operator for the indexer. Not sure how complex to implement this would be, if it's too much then don't worry for now.

src/mappings/subgraphService.ts
juanmardefago: Good catch, and yes, it's incorrect

juanmardefago: I think I added those helper functions in a separate PR though, will have to double check.

Another thing I was thinking is that updating everything PR by PR will be a rebase nightmare (as each PR depends on the previous, so a lot of rebases going to happen. I might simply aggregate all these changes, merge everything as is, and then add a bunch of PRs to fix each review separately 😓

schema.graphql
juanmardefago: I think we tracked forced closures in another entity before, so it's definitely doable.
Same for the legacy flag.

schema.graphql
juanmardefago: Good idea

src/mappings/subgraphService.ts
tmigone: absolutely agree with putting everything into one big PR

schema.graphql
tmigone: I changed the `AllocationClosed` event to also emit a boolean that specifies wether the allocation was force closed or not.

```
    event AllocationClosed(
        address indexed indexer,
        address indexed allocationId,
        bytes32 indexed subgraphDeploymentId,
        uint256 tokens,
        bool forceClosed
    );
```

juanmardefago: @tmigone which branch of the contracts repo did you add this to? Trying to generate a new ABI for the subgraph to use with this change.

juanmardefago: Added it with the same checks we've been already doing for a while, but if the event signature changed, we will need to change it too, so whenever you get a chance to look at this let me know so I can update the latest ABI 😅

src/mappings/subgraphService.ts
juanmardefago: Actually no need to calculate, but rather use the actual latest calculated value from the provision 😅 .

Also thanks to this I realized the fields we were using for the AllocationClosed case were the same as the ones we were using for the AllocationCreated case (i.e. we were using `indexingRewardEffectiveCutAtStart` on both, instead of `indexingRewardEffectiveCutAtStart` for Created, and `indexingRewardEffectiveCutAtClose` for Closed 😅)

PR #286, opened 2025-02-25
[Horizon] PaymentCollection implementation

juanmardefago:


schema.graphql
```diff
@@ -900,6 +901,22 @@ type Provision @entity {
...
+  "The total amount of indexing rewards the indexer kept"
+  indexerIndexingRewards: BigInt!
+  "The total amount of indexing rewards given to delegators"
+  delegatorIndexingRewards: BigInt!
+
+  # Query fees
+  "Total query fees collected. Includes the portion given to delegators"
+  queryFeesCollected: BigInt!
+  "Query fee amount collected from the protocol. Does not include portion given to delegators"
+  providerQueryFees: BigInt!
```
tmigone, line 916:
```suggestion
  indexerQueryFees: BigInt!
```

just for consistency since we used `indexerIndexingRewards` above

src/mappings/graphPayments.ts
```diff
@@ -0,0 +1,25 @@
...
+    )
+    epoch.taxedQueryFees = epoch.taxedQueryFees.plus(event.params.tokensProtocol)
+    epoch.save()
+
+    // update graph network
+    graphNetwork.totalTaxedQueryFees = graphNetwork.totalTaxedQueryFees.plus(event.params.tokensProtocol)
+    graphNetwork.save()
+
+    // Replicate for payment source specific data
+    let paymentSource = createOrLoadPaymentSource(event.params.payer)
```
tmigone, line 20: Worth noting here that `event.params.payer` is the caller of `GraphPayments.collect()`. In most cases this will be the `PaymentsEscrow` contract. Not to be confused with the "payer" entity as the one that originally provided the funds to the `PaymentsEscrow`. 

I don't think code here is incorrect, just noting that it could be misleading.

juanmardefago: Added a small comment for this, might revisit this later as I don't know exactly how to make it work like it used to in V1. cc: @tmigone

PR #287, opened 2025-03-03
[Horizon] Delegation support

juanmardefago:


schema.graphql
```diff
@@ -893,12 +893,22 @@ type Provision @entity {
...
 
+  "Raw indexing rewards cut. In Horizon this is amount of indexing rewards that the delegation pool receives in PPM"
   indexingRewardsCut: BigInt!
 
+  "The percent of indexing rewards generated by the delegated stake that the Indexer keeps for itself"
+  indexingRewardEffectiveCut: BigDecimal!
+  "The percent of query rebate rewards generated by the delegated stake that the Indexer keeps for itself"
+  queryFeeEffectiveCut: BigDecimal!
+  "The percent of reward dilution delegators experience because of overdelegation. Overdelegated stake can't be used to generate rewards but still gets accounted while distributing the generated rewards. This causes dilution of the rewards for the rest of the pool."
+  overDelegationDilution: BigDecimal!
```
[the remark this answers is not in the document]
tmigone: I thiink having a numeric example here would be nice, similar to what you did below with `indexerRewardsOwnGenerationRatio`

schema.graphql
```diff
@@ -924,6 +934,12 @@ type Provision @entity {
   delegatorShares: BigInt!
   "Exchange rate of of tokens received for each share"
   delegationExchangeRate: BigDecimal!
+  "Ratio between the amount of the indexers own stake over the total usable stake."
+  ownStakeRatio: BigDecimal!
+  "Ratio between the amount of delegated stake over the total usable stake."
+  delegatedStakeRatio: BigDecimal!
+  "Percentage of indexers' own rewards received in relation to its own stake. 1 (100%) means that the indexer is receiving the exact amount that is generated by his own stake"
+  indexerRewardsOwnGenerationRatio: BigDecimal!
```
tmigone, line 942: Curious why all these new variables? Is it to make the APY calculations easier?

tmigone, line 942: Maybe worth clarifying this ratio can be below or above 1 depending on the delegation fee cuts.

src/mappings/helpers/helpers.ts
```diff
@@ -916,6 +961,68 @@ export function updateAdvancedIndexerMetrics(indexer: Indexer): Indexer {
...
+    ? BigDecimal.fromString('0')
+    : BigDecimal.fromString('1') - delegatorCut / provision.delegatedStakeRatio
+}
+
+export function calculateIndexerRewardOwnGenerationRatioForProvision(provision: Provision): BigDecimal {
+  let rewardCut =
+    provision.indexingRewardsCut.toBigDecimal() / BigDecimal.fromString('1000000')
+  return provision.ownStakeRatio == BigDecimal.fromString('0')
+    ? BigDecimal.fromString('0')
+    : rewardCut / provision.ownStakeRatio
```
tmigone, line 1000:
```suggestion
    : (BigDecimal.fromString('1') - rewardCut) / provision.ownStakeRatio
```

I think this is the correct calculation. What we are looking for (as far as I understood from the schema definition), is the ratio between the `total query fees obtained by the indexer` and the `total query fees generated by the indexer's stake`. A ratio > 1 indicates the indexer is getting some fees from the delegator's stake, a ratio < 1 the opposite. 

Is this correct?

schema.graphql
tmigone, line 942: Are we not interested in `queryFeesOwnGenerationRatio`?

src/mappings/helpers/helpers.ts
```diff
@@ -916,6 +961,68 @@ export function updateAdvancedIndexerMetrics(indexer: Indexer): Indexer {
...
+}
+
+export function calculateOverdelegationDilutionForProvision(provision: Provision): BigDecimal {
+  let provisionedTokensBD = provision.tokensProvisioned.toBigDecimal()
+  let delegatedTokensBD = provision.delegatedTokens.toBigDecimal()
+  let graphNetwork = GraphNetwork.load('1')!
+  let delegationRatioBD = BigInt.fromI32(graphNetwork.delegationRatio).toBigDecimal()
+  let maxDelegatedStake = provisionedTokensBD * delegationRatioBD
```
tmigone, line 1008: I think this calculation is incorrect. The max useful delegated stake would be the provisioned tokens times the delegation ratio of the target data service. 

The delegation ratio is stored in the data service so we would need to query it from there, which probably means adding a `delegationRatio` property to DataService entity? When the delegation ratio is set an event is emitted on the data service contract: `event DelegationRatioSet(uint32 ratio)`.

Worth noting as well that delegation ratio is NOT MANDATORY, meaning a data service might consider all delegation as useful stake and not even have a delegation ratio. So we should probably initialize the delegationRatio as MAX uint256?

Let me know if you want to discuss this.

src/mappings/horizonStaking.ts
```diff
@@ -279,4 +279,184 @@ export function handleTokensToDelegationPoolAdded(event: TokensToDelegationPoolA
...
+        if (delegator.activeStakesCount == 1) {
+            graphNetwork.activeDelegatorCount = graphNetwork.activeDelegatorCount + 1
+        }
+    }
+
+    graphNetwork.save()
+    delegator.save()
+}
+
+export function handleDelegationSlashed(event: DelegationSlashed): void {
```
tmigone, line 363: We should probably add `tokensSlashedServiceProvider` and `tokensSlashedDelegationPool` to the `Provision` entity to track those amounts.

schema.graphql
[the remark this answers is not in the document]
juanmardefago: This has actually been in the original version for quite some time, we just ported all "advanced" metrics for the schema to be cohesive with the previous version, since essentially, for Horizon (stage 1 at least) a Provision entity is almost exactly the same thing as the Indexer entity was on the original subgraph. (A provision is the view of the indexer for a specific data service, and if we consider only the subgraph service, the only thing it adds is the the provision <> deprovision flows)

schema.graphql
juanmardefago: Just advanced metrics we had before for the indexer, as mentioned in the previous comment

juanmardefago: To be honest, all these advanced metrics were implemented really back in the day when we still used spreadsheets to figure out potentially interesting metrics that could tell us how indexers behaved on a protocol that only had indexing rewards.

I doubt people are using them any more, but even so, we might want to eventually remove the advanced metrics and keep things simple, as the approach of keep adding more and more advanced stuff is what lead us to this state where we have a huge amount of fields that all/most people won't ever use and that require porting or maintenance every so often (during upgrades for example)

src/mappings/helpers/helpers.ts
juanmardefago: yeah, I think I missed updating this one, since all previous ones were updated (and changed the naming from `rewardCut` to `delegatorCut`, but this one wasn't 😅

src/mappings/helpers/helpers.ts
juanmardefago: Yeah, seems this one also slipped through the cracks, and the clarification makes sense, but the metric itself lacks any point if the data service allows all stake to be used, since that means there's no dilution (i.e. there can never be any extra delegation taking rewards/fees that they aren't helping generate, since useful delegation isn't capped on that case)

schema.graphql
```diff
@@ -924,6 +934,12 @@ type Provision @entity {
   delegatorShares: BigInt!
   "Exchange rate of of tokens received for each share"
   delegationExchangeRate: BigDecimal!
+  "Ratio between the amount of the indexers own stake over the total usable stake."
```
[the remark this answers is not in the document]
juanmardefago: Usable stake implies that it's capped by delegation ratio where it applies though.

schema.graphql
juanmardefago: I'd rather not add more extra advanced metrics, but just comply with the existing ones for now

PR #289, opened 2025-04-26
feat: added HorizonDisputeManager

juanmardefago:


schema.graphql
```diff
@@ -219,8 +219,12 @@ type GraphNetwork @entity {
   slashingPercentage: Int!
   "Minimum deposit to create a dispute"
   minimumDisputeDeposit: BigInt!
-  "Reward to Fisherman on successful disputes. In parts per million"
+  "[LEGACY] Reward to Fisherman on successful disputes. In parts per million"
   fishermanRewardPercentage: Int!
+  "[HORIZON]Reward to Fisherman on successful disputes. In parts per million"
+  fishermanRewardCut: Int!
+  "Maximum Penalty to Indexer on successful disputes for indexing disputes. In parts per million"
```
tmigone, line 226: Might be worth saying this is a percentage of the indexer's stake snapshot (own stake plus delegated).

schema.graphql
tmigone, line 1: I think it's worth adding a `isLegacy` boolean to the `Disputes` entity like we did for allocations and other stuff.

PR #290, opened 2025-05-08
feat: invert cuts to keep meaning consistent with pre-horizon

tmigone:


src/mappings/staking.ts
```diff
@@ -51,19 +51,6 @@ import {
...
-  indexer.indexingRewardCut = event.params.indexingRewardCut.toI32()
-  indexer.queryFeeCut = event.params.queryFeeCut.toI32()
-  indexer.delegatorParameterCooldown = event.params.cooldownBlocks.toI32()
-  indexer.lastDelegationParameterUpdate = (
-    addresses.isL1 ? event.block.number : graphNetwork.currentL1BlockNumber!
-  ).toI32()
-  indexer = updateAdvancedIndexerMetrics(indexer as Indexer)
-  indexer.save()
-}
-
```
juanmardefago, line 66: I would not delete this at all, since we essentially lose historical tracking of delegation cuts. Particularly given that we are now moving all of the new delegation cuts to line up with these ones 😅

src/mappings/helpers/helpers.ts
```diff
@@ -226,9 +226,9 @@ export function createOrLoadProvision(indexerAddress: Bytes, verifierAddress: By
     provision.maxVerifierCutPending = BigInt.fromI32(0)
     provision.thawingPeriod = BigInt.fromI32(0)
     provision.thawingPeriodPending = BigInt.fromI32(0)
-    provision.queryFeeCut = BigInt.fromI32(0)
-    provision.indexingFeeCut = BigInt.fromI32(0)
-    provision.indexingRewardsCut = BigInt.fromI32(0)
+    provision.queryFeeCut = BigInt.fromI32(1000000)
+    provision.indexingFeeCut = BigInt.fromI32(1000000)
+    provision.indexingRewardsCut = BigInt.fromI32(1000000)
```
juanmardefago, line 231: Do we want to move all delegation cuts to the legacy version? Or do we keep Provision level ones lined up with Horizon, and Indexer level ones lined up with Legacy?

I think the latter is better, since it allows the UIs to move to "Stage 2"-like cuts or keep Legacy cuts by using different entities, and keeping both as Legacy just means that once we move to a more generic network subgraph (Stage 2) we'll have extra work that needs to be timed correctly

juanmardefago:
@tmigone this should be in line with the discussions we had around this topic last time.
Legacy cuts should obey legacy logic (cut that the indexer takes), and are now all prefixed with `legacy`. Stuff that depended on that value should now use the legacy one for legacy code (i.e. on al non-horizon files), as well as the need for a full refactor of advanced metrics on legacy code, since there was also the effective cuts, and some advanced metrics that needed to be split into legacy vs horizon versions (otherwise they wouldn't have made sense)

PR #293, opened 2025-05-16
[Horizon] Stage 1 subgraph upgrade

juanmardefago:


juanmardefago:
Manually reverted this since it was mistakenly merged by yours truly.

PR #296, opened 2025-05-16
[Horizon] Stage 1 upgrade

juanmardefago:


src/mappings/graphTallyCollector.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export function handlePaymentCollected(event: PaymentCollected): void {
+    let index = event.logIndex.toI32()
+    let transactionId = event.transaction.hash.concatI32(index)
+    let transaction = new PaymentsEscrowTransaction(transactionId)
+
+    let payer = createOrLoadPayer(event.params.payer)
+    let receiver = createOrLoadReceiver(event.params.receiver)
+    let escrow = createOrLoadEscrowAccount(event.params.payer, event.params.dataService, event.params.receiver)
+    escrow.balance = escrow.balance.minus(event.params.tokens)
```
tmigone, line 54: I think we should update the escrow as part of the payments escrow handlers, specifically in `EscrowCollected` handler as that is where the actual state is changed.

tmigone: And we should add `tokensCollected` to the graph tally global state

src/mappings/graphTallyCollector.ts
```diff
@@ -50,8 +50,7 @@ export function handlePaymentCollected(event: PaymentCollected): void {
 
     let payer = createOrLoadPayer(event.params.payer)
     let receiver = createOrLoadReceiver(event.params.receiver)
-    let escrow = createOrLoadEscrowAccount(event.params.payer, event.params.dataService, event.params.receiver)
-    escrow.balance = escrow.balance.minus(event.params.tokens)
+    let escrow = createOrLoadEscrowAccount(event.params.payer, event.transaction.to, event.params.receiver)
```
tmigone, line 53: `event.transaction.to` is the collector here no?

tmigone, line 53: I think we dont need to instance the escrow account in this handler now no?

Maikol: We need it to link the escrow account with the transaction.

PR #299, opened 2025-05-19
feat: adding latest value for thawingUntil on Indexer and Provision

juanmardefago:


tmigone (approved):

juanmardefago:
@tmigone changed the logic so that it's always the max value

src/mappings/horizonStaking.ts
```diff
@@ -270,6 +270,14 @@ export function handleThawRequestCreated(event: ThawRequestCreated): void {
         throw new Error("Invalid thaw request type")
     }
     request.save()
+
+    // update latest thawingUntil for provision and indexer
+    let provision = createOrLoadProvision(event.params.serviceProvider, event.params.verifier, event.block.timestamp)
+    provision.thawingUntil = event.params.thawingUntil > provision.thawingUntil ? event.params.thawingUntil : provision.thawingUntil
+    provision.save()
+
+    indexer.thawingUntil = event.params.thawingUntil > indexer.thawingUntil ? event.params.thawingUntil : indexer.thawingUntil
```
andrascodes, line 279: Sry I was just reviewing this more closely for the Engineering Plan: don't we have to set `indexer.thawingUntil` back to 0 in `handleThawRequestFulfilled` or `handleTokensDeprovisioned`?

andrascodes: yeah actually you're right! shouldn't be a big deal! It's overwritten by a later timestamp anyway and if it's in the past it's withdrawable as you say. nvm then! thx!!!

juanmardefago: I'm gonna merge it, but worst case scenario we can modify it later down the road if necessary

PR #312, opened 2025-11-03
fix: staking legacy contract call when slashing

Maikol:


src/mappings/staking.ts
```diff
@@ -273,9 +289,7 @@ export function handleStakeDelegatedLocked(event: StakeDelegatedLocked): void {
 
   delegatedStake.unstakedTokens = delegatedStake.unstakedTokens.plus(event.params.tokens)
   delegatedStake.shareAmount = delegatedStake.shareAmount.minus(event.params.shares)
-  delegatedStake.lockedTokens = delegatedStake.lockedTokens.plus(event.params.tokens)
```
juanmardefago: Same as mentioned before, repeats for the remaining value updates removed

